PR #1, opened 2021-06-06
Refactor ros2_ws image and scripts

eeberhard:
* Add sshd_config file to Dockerfile

* Streamline build.sh script and support MacOS

* Add server.sh script with arg parsing to spinup
a built image as a background SSH server.

* Add connect.sh script as a shortcut for 'exec'
attaching an interactive shell to a running
container, with platform specific environment
variables and volumes for display configuration.

* Add apt-key file downloaded from
http://repo.ros2.org/repos.key
to ros2_ws/config directory to add with apt-key in
the Dockerfile. Note: curl is not used because
it is not yet installed in the base ros image,
and cannot be installed with apt without the key!

* Consolidate contents of ros2_ws config directory into
Dockerfile directly (99_aptget and update_bashrc),
both to reduce clutter and to allow passing the USER
variable instead of hardcoding ros2

* Reduce the amount of setup sourcing done in the various
bashrc and entrypoint scripts in the Dockerfile, simply
sourcing the core ros and ros2_ws setup scripts at the top
of bashrc

--- 

Before I forget what I did, here are the changes I mainly made to enable CLion development. At the same time, I've tried to make the docker image namespace more consistent with the conventions I set out a while ago, and also found a convenient solution to the whole UID and GID thing on linux.

The idea for the workflow is to instantiate a given ros2 container only once, and have it run as a background process with an SSH server. This runs non-interactively as root, so on instantiation, it can copy the RSA authentication as well as change the UID and GID of a specified user (ros2) to match that of the host user. Then, subsequent sessions can use `docker exec` to attach to the existing container (rather than creating a new one each time).

---

How does this enable the CLion workflow? Because the ros2 sourcing happens at the top of the bashrc file, it is now evaluated even for non-interactive sessions (e.g. the CLion client ssh process). Add to this the ability to keep the ros2 user passwordless like before and you have a happy situation on your hands. There's some extra CMake magic to really get the full IDE experience with ros2 going, but at least in its current form you can load a source project and the colcon generated cmakelist and it will happily index it.


buschbapti:
This looks super interesting and promising. It builds like a charm, the `server.sh` script correctly create the container, however on `connect.sh` I got:

```
access control disabled, clients can connect from any host
unknown shorthand flag: 'v' in -v
See 'docker container exec --help'.
```

buschbapti:
The volume option does not appear to be an option for `docker exec` but for `docker run`

buschbapti (approved):

PR #2, opened 2021-07-12
Feature/general scripts

eeberhard:
I made some changes to make the scripts more generic so that they are easy to re-use across different images. I also added a ROS (1) workspace image.

The ros_ws is a near copy of the ros2_ws, but invokes catkin_make instead of colcon build and has a ros user instead of ros2. There is some code duplication, but because of docker build contexts it is difficult to reuse elements between the Dockerfiles.

### DIFF
--- a/ros_ws/Dockerfile
+++ b/ros_ws/Dockerfile
@@ -1,0 +1,85 @@
+ARG ROS_DISTRO=noetic
+FROM ros:${ROS_DISTRO} as base-dependencies
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt update && apt install -y \
+  autoconf \
+  automake \
+  build-essential \
+  curl \
+  g++ \
+  gdb \
+  iputils-ping \
+  libboost-all-dev \
+  libeigen3-dev \
+  libtool \
+  libssl-dev \
+  make \
+  mesa-utils \
+  python3-setuptools \
+  python3-pip \
+  python3-rosdep \
+  python3-rosinstall \
+  python3-rosinstall-generator \
+  python3-wstool \
+  ros-${ROS_DISTRO}-rviz \
+  rsync \
+  software-properties-common \
+  ssh \
+  sudo \
+  unzip \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN sudo rosdep update
+
+RUN echo "Set disable_coredump false" >> /etc/sudo.conf
+
+# Configure sshd server settings
+RUN ( \
+    echo 'LogLevel DEBUG2'; \
+    echo 'PubkeyAuthentication yes'; \
+    echo 'Subsystem sftp /usr/lib/openssh/sftp-server'; \
+  ) > /etc/ssh/sshd_config_development \
+  && mkdir /run/sshd
+
+
+FROM base-dependencies as base-workspace
+ENV USER ros
+ENV HOME /home/${USER}
+
+# create amd configure a new user
+ARG UID=1000
+ARG GID=1000
+RUN addgroup --gid ${GID} ${USER}
+RUN adduser --gecos "ROS User" --uid ${UID} --gid ${GID} ${USER} && yes | passwd ${USER}
+RUN usermod -a -G dialout ${USER}
+RUN echo "${USER} ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers.d/99_aptget
+RUN chmod 0440 /etc/sudoers.d/99_aptget && chown root:root /etc/sudoers.d/99_aptget
+
+# Configure sshd entrypoint to authorise the new user for ssh access and
+# optionally update UID and GID when invoking the container with the entrypoint script
+COPY ./config/sshd_entrypoint.sh /sshd_entrypoint.sh
+RUN chmod 744 /sshd_entrypoint.sh
+
+# build ROS workspace
+USER ${USER}
+WORKDIR ${HOME}/ros_ws/
+RUN mkdir -p src
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib
+RUN /bin/bash -c "source /opt/ros/${ROS_DISTRO}/setup.bash; catkin_make"
+
+# set up environment
+USER root
+
+# prepend the environment sourcing to bashrc (appending will fail for non-interactive sessions)
+RUN echo "source /opt/ros/${ROS_DISTRO}/setup.bash; \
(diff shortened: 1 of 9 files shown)


domire8:
One question I have here: up until now we always had the `update_bashrc` and `ros_entrypoint.sh` scripts to complete the build of a ros ws. I see that `update_bashrc` has been replaced with two RUN commands. But what about `ros_entrypoint.sh` ? Is that not needed? Or was it never actually needed?  I'm not even sure what it really is for...

domire8:
Also, what is the difference between pulling `FROM ros:${ROS_DISTRO} as base-dependencies` and `FROM osrf/ros:noetic-desktop` ?

ros_ws/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+  libssl-dev \
+  make \
+  mesa-utils \
+  python3-setuptools \
+  python3-pip \
+  python3-rosdep \
+  python3-rosinstall \
+  python3-rosinstall-generator \
+  python3-wstool \
+  ros-${ROS_DISTRO}-rviz \
```
domire8, line 25: What ros packages are preinstalled with this image? I'm thinking mainly about ros-control and ros-controllers...
Also, I would propose to add `ros-distro-joint-state-publisher-gui` here if it's not installed already because I find this is quite an important package and should be there by default.

ros_ws/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+
+# set up environment
+USER root
+
+# prepend the environment sourcing to bashrc (appending will fail for non-interactive sessions)
+RUN echo "source /opt/ros/${ROS_DISTRO}/setup.bash; \
+source /home/${USER}/ros_ws/devel/setup.bash" | cat - ${HOME}/.bashrc > tmp && mv tmp ${HOME}/.bashrc
+RUN echo "session required pam_limits.so" | sudo tee --append /etc/pam.d/common-session > /dev/null
+
+WORKDIR ${HOME}
```
domire8, line 79: I like to set the WORKDIR to ${HOME}/ros_ws in "my" images because it saves you the `cd ros_ws` at the beginning and given that this is mainly a ros workspacce there is no reason there should be other files in ${HOME}. What do you think about that?

ros_ws/Dockerfile
domire8: Ah nevermind, I just realized from your example that this is really just a base image and you will install whatever you want in the Dockerfile in your project...

scripts/run_interactive.sh
```diff
@@ -0,0 +1,74 @@
...
+if [ -n "$USERNAME" ]; then
+  RUN_FLAGS+=(-u "${USERNAME}")
+fi
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  RUN_FLAGS+=(-e DISPLAY=host.docker.internal:0)
+else
+  xhost +
+  RUN_FLAGS+=(-e DISPLAY="${DISPLAY}")
+  RUN_FLAGS+=(-e XAUTHORITY="${XAUTH}")
```
domire8, line 67:
```suggestion
  RUN_FLAGS+=(-e XAUTHORITY="${XAUTHORITY}")
  RUN_FLAGS+=(-v "/tmp/.X11-unix:/tmp/.X11-unix:rw")
```

domire8 (commented):
I did a few tests and without the volume, I cannot launch rviz because of display issues

scripts/run_interactive.sh
eeberhard: Do we also want `RUN_FLAGS+=(--volume="${XAUTHORITY}:${XAUTHORITY}")`?

ros_ws/Dockerfile
eeberhard: yeah I thought about this, and I did it in my derived example image. might as well do it here as well 👍

eeberhard:
> One question I have here: up until now we always had the update_bashrc and ros_entrypoint.sh scripts to complete the build of a ros ws

I found that the ros entrypoint script is no longer necessary. If you look at what it does, it sources the bashrc, sets a starting directory and forwards all other command arguments to bash.

- bashrc is automatically invoked at login anyway (with the default login CMD)
- We can already set the starting directory with WORKDIR
- If we want to provide extra command arguments to bash, we can invoke `/bin/bash ...` command ourselves on run or exec. Similarly the entrypoint can be manually set with the command override (like with the sshd script)

Derived images are welcome to add their own entrypoint scripts if they need / want more customisation. For the base image I wanted to trim down the extra files and options to the minimum.

scripts/run_interactive.sh
domire8: I am not sure to be honest. I added it and I couldn't see a difference when I had it and when I didn't.

domire8:
> > One question I have here: up until now we always had the update_bashrc and ros_entrypoint.sh scripts to complete the build of a ros ws
> 
> I found that the ros entrypoint script is no longer necessary. If you look at what it does, it sources the bashrc, sets a starting directory and forwards all other command arguments to bash.
> 
>     * bashrc is automatically invoked at login anyway (with the default login CMD)
> 
>     * We can already set the starting directory with WORKDIR
> 
>     * If we want to provide extra command arguments to bash, we can invoke `/bin/bash ...` command ourselves on run or exec. Similarly the entrypoint can be manually set with the command override (like with the sshd script)
> 
> 
> Derived images are welcome to add their own entrypoint scripts if they need / want more customisation. For the base image I wanted to trim down the extra files and options to the minimum.

Fair enough, I think you did a great job here, the scripts are very helpful like that.

buschbapti (approved):
This is all good. I suggest we actually add the one for `rviz`, `rviz2` and `bridge`. This can be a separate PR if you think it is out of scope / gets too big

domire8 (approved):
The scripts are very nice and I think this can be merged like that.
However, we all have errors when trying to launch rviz so it would be nice if this could be fixed with the next PR, maybe by using the `osrf/ros` images

ros_ws/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+    echo 'Subsystem sftp /usr/lib/openssh/sftp-server'; \
+  ) > /etc/ssh/sshd_config_development \
+  && mkdir /run/sshd
+
+
+FROM base-dependencies as base-workspace
+ENV USER ros
+ENV HOME /home/${USER}
+
+# create amd configure a new user
```
domire8, line 50:
```suggestion
# create and configure a new user
```

eeberhard:
I agree that a second PR is necessary to fix some of the settings for rviz, as well as one to bring in more images as tools. The main focus of this PR was to revise the scripts (while adding a copied ros image), so I will merge this one now to keep things manageable.

PR #3, opened 2021-08-02
Fix permission issues without Docker BuildKit

domire8:
So I was trying to build the images without `DOCKER_BUILDKIT=1` in the build scripts and I ran into some permission issues. My interpretation is that `WORKDIR ros_ws` creates and sets the new directory but without the buildkit, it has the wrong permissions, so the user is not able to create another directory there. By creating the directory before setting the WORKDIR, this issue can be avoided and the image can be built without the buildkit. 
I was trying that out because of the CI that we want to do and github actions don't know the buildkit. What do you think?

### DIFF
--- a/ros2_ws/Dockerfile
+++ b/ros2_ws/Dockerfile
@@ -5,7 +5,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 COPY config/apt-key /tmp/apt-key
 RUN apt-key add /tmp/apt-key
 
-RUN apt update && apt install -y \
+RUN apt-get update && apt-get install -y \
   autoconf \
   automake \
   build-essential \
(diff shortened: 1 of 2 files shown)


ros2_ws/Dockerfile
```diff
@@ -5,7 +5,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 COPY config/apt-key /tmp/apt-key
 RUN apt-key add /tmp/apt-key
 
-RUN apt update && apt install -y \
+RUN apt-get update && apt-get install -y \
```
eeberhard, line 8: Any justification for this change? apt and apt-get are equivalent for these commands (apt being a wrapper for apt-get)

eeberhard (approved):
Nice catch with WORKDIR not respecting USER (as shown in [the link you sent me](https://github.com/moby/moby/issues/20295)). Creating the directory in advance is a good idea.

I see also there would also be the option of adding `WORKDIR --chown ${USER} ...` in other cases. In this case the `mkdir` is preferable because we also need the src folder on top.

ros2_ws/Dockerfile
domire8: I always get a warning `WARNING: apt does not have a stable CLI interface. Use with caution in scripts.` when using apt instead of apt-get

PR #4, opened 2021-08-03
Build and push ROS images in CI

domire8:
This PR adds two workflows, one to build and push the noetic and foxy image upon push on main and one that can be manually dispatched with a desired configuration as input (e.g. starting the workflow with `aica-technology/ros2_ws:galactic` will build a ros2 galactic image). It uses the `secrects.GITHUB_TOKEN` to have write and read permissions, so no other secret is needed. It will also automatically connect the packages to the repository (obviously, as it is now the repo that creates and pushes the package and not the user).

### DIFF
--- a/.github/actions/build-push/action.yml
+++ b/.github/actions/build-push/action.yml
@@ -19,7 +19,7 @@ runs:
         ROS_DISTRO=${IMAGE_NAME#*:}
         WORKSPACE=${IMAGE_ID#*/}
         WORKSPACE=${WORKSPACE/-/_}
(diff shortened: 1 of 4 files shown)


domire8:
See my fork (that I'm gonna delete afterwards) https://github.com/domire8/docker-images, where this is already working

eeberhard (approved):
Looks pretty clean to me! Nice syntax with the input parsing to split between the ros and ros2 folders, so that we can have the single action file for both. The tests on your fork also look good.

One thing, did you want to add the LABEL command to the Dockerfiles? I don't think it's needed since it's already associated with the repo (or can be manually linked after creation).

Thanks for adding this, looking forward to seeing it run once merged into the default branch!

domire8:
> One thing, did you want to add the LABEL command to the Dockerfiles? I don't think it's needed since it's already associated with the repo (or can be manually linked after creation).

I copied this a bit, I guess it's not really useful after all.

PR #6, opened 2021-08-05
Install git

domire8:
I'm sorry to open a PR for that but I kinda think git is quite essential :smile:

### DIFF
--- a/ros2_ws/Dockerfile
+++ b/ros2_ws/Dockerfile
@@ -12,6 +12,7 @@ RUN apt-get update && apt-get install -y \
   curl \
   g++ \
   gdb \
+  git \
   iputils-ping \
   libboost-all-dev \
   libeigen3-dev \
--- a/ros2_ws/Dockerfile
+++ b/ros2_ws/Dockerfile
@@ -8,26 +8,19 @@ RUN apt-key add /tmp/apt-key
 RUN apt-get update && apt-get install -y \
   autoconf \
   automake \
-  build-essential \
   curl \
-  g++ \
   gdb \
   git \
   iputils-ping \
   libboost-all-dev \
-  libeigen3-dev \
   libtool \
-  libssl-dev \
-  make \
   mesa-utils \
   nano \
-  python3-setuptools \
   python3-pip \
   ros-${ROS_DISTRO}-rviz2 \
   rsync \
   software-properties-common \
   ssh \
-  sudo \
   unzip \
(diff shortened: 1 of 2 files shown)


ros2_ws/Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN apt-get update && apt-get install -y \
   curl \
   g++ \
   gdb \
+  git \
```
eeberhard, line 15: From the VNC tests I did last week, we only need to specify the following additional packages (everything else is already in the base image). Saves a few lines and bit of build time, not worth doing on its own but if you're already making a PR it would fit.
```
  curl \
  gdb \
  git \
  iputils-ping \
  libeigen3-dev \
  libtool \
  mesa-utils \
  nano \
  python3-pip \
  python3-rosinstall-generator \
  ros-${ROS_DISTRO}-rviz \
  rsync \
  software-properties-common \
  ssh \
  unzip \
```

domire8: Does that hold for ros_ws too?

eeberhard: This comment actually applies to the ros_ws, not ros2_ws. For ros2 it would be:

```
  autoconf \ <-- (maybe, might not be needed)
  automake \ <-- (maybe, might not be needed)
  curl \
  gdb \
  git \
  iputils-ping \
  libboost-all-dev \ <-- (maybe, might not be needed)
  libtool \
  mesa-utils \
  nano \
  python3-pip \
  ros-${ROS_DISTRO}-rviz2 \
  rsync \
  software-properties-common \
  ssh \
  unzip \
```

eeberhard: In the end I guess it's probably safer to double up. But, it might be nice to try some downstream projects with the reduced set here just so we save installing stuff we don't need (like lib boost on ros2 base)

eeberhard (approved):

ros2_ws/Dockerfile
domire8: Okay, I kept the *maybes*. If you prefer to take them out, just make a suggestion on the PR and we're good to go. I don't mind.

eeberhard: let's keep the maybes, better to be safe for now (since we're not currently worried about image size or build time in our applications, not doing a huge amount in CI)

eeberhard (approved):

PR #7, opened 2021-08-09
Add GPU option

buschbapti:
This PR adds the `--gpus` option to the `run_interactive.sh` script. It basically forward the option to the docker engine to respect docker options for GPU usage. Most common usage will be `--gpus all`.

When the `--gpus` flag is set, it also set the `nvidia` environment variables. This avoids having to set them in the `Dockerfile`.

### DIFF
--- a/scripts/run_interactive.sh
+++ b/scripts/run_interactive.sh
@@ -3,6 +3,7 @@
 IMAGE_NAME=""
 CONTAINER_NAME=""
 USERNAME=""
+GPUS=""
 
 HELP_MESSAGE="
 Usage: ./run_interactive.sh <image> [-n <name>] [-u <user>]
@@ -28,6 +29,11 @@ Options:
                            mount between the host and container
                            as two full paths separated by a ':'.
 
+  --gpus <gpus_options>    Add GPU access for applications that
+                           require hardware acceleration (e.g. Gazebo)
+                           For the list of gpus_options parameters
+                           see https://docs.docker.com/config/containers/resource_constraints/
+
   -h, --help               Show this help message."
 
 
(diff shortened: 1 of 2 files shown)


scripts/server.sh
```diff
@@ -44,6 +45,11 @@ Options:
   -u, --user <user>        Specify the name of the remote user.
                            (default: ${USERNAME})
 
+  --gpus <gpus_options>    Add GPU access for applications that
+                           require hardware acceleration (e.g. Gazebo)
+                           For the list of gpus_options parameters
+                           see https://docs.docker.com/config/containers/resource_constraints/
```
domire8, line 51:
```suggestion
  --gpus <gpu_options>     Add GPU access for applications that
                           require hardware acceleration (e.g. Gazebo)
                           For the list of gpu_options parameters
                           see https://docs.docker.com/config/containers/resource_constraints/
```

scripts/run_interactive.sh
```diff
@@ -28,6 +29,11 @@ Options:
                            mount between the host and container
                            as two full paths separated by a ':'.
 
+  --gpus <gpus_options>    Add GPU access for applications that
+                           require hardware acceleration (e.g. Gazebo)
+                           For the list of gpus_options parameters
+                           see https://docs.docker.com/config/containers/resource_constraints/
```
domire8, line 35:
```suggestion
  --gpus <gpu_options>     Add GPU access for applications that
                           require hardware acceleration (e.g. Gazebo)
                           For the list of gpu_options parameters
                           see https://docs.docker.com/config/containers/resource_constraints/
```

domire8 (approved):
Looks good to me! Another improvement done

PR #10, opened 2021-08-31
Common sshd entrypoint

eeberhard:
As @domire8 mentioned this morning, apparently @hubernikus was interested in using the SSH and `aica-docker server` approach on a docker image that is not based on the workspaces we provide. In fact, this functionality is provided by a custom entrypoint script that could be a part of any image. In addition, I was already unhappy that the script was duplicated across two workspaces (because the docker build context is limited, and can't follow symlinks).

In this PR, I move the sshd_entrypoint script to a common directory with README instructions on usage. I then modified the build scripts for the workspaces to make a local copy of the file so that it resides within the build context. I also made the server script more robust by always running `docker run --user root ...`, so that a Dockerfile doesn't have to end with `USER root`. Finally, I reformatted the scripts (the final commit d8668bb does not change any logic).

All in all, this change reduces code duplication and should make it much easier for other users to integrate `aica-docker server` into their lives.


domire8:
Well isn't that just great!!! Nice work. @buschbapti won't like it because he's not patient enough to pull the ros workspace images all the time :smile:

domire8 (approved):
Well isn't that just great!!! Nice work. @buschbapti won't like it because he's not patient enough to pull the ros workspace images all the time :smile:

hubernikus:
Suuper nice! Thanks a lot. I still struggle a bit with the Dockerfile syntax, but seems very legit :sweat_smile:

hubernikus (approved):

PR #11, opened 2021-09-01
Correct the multiple occurences replacement

buschbapti:
In the server script, the container name is created by replacing the `/` in the image name by `-`. The problem is this only does it for the first occurrence. Adding an extra '/' in the command corrects this issue.

### DIFF
--- a/scripts/src/interactive.sh
+++ b/scripts/src/interactive.sh
@@ -88,7 +88,7 @@ if [ -z "${IMAGE_NAME}" ]; then
 fi
 
 if [ -z "${CONTAINER_NAME}" ]; then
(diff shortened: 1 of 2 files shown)


domire8:
Does it now catch all occurrences or only the first two?

buschbapti:
You made me doubt but it does catch all of them

domire8 (approved):

eeberhard:
Nice catch. Does this also apply to the interactive script? If so please make the change there also.

It's also necessary to catch and parse any `.`, such as in `ghcr.io/...`.

buschbapti:
Thanks for checking. It is also added in the `interactive` now and handles the `.` as requested.

domire8 (approved):

PR #13, opened 2021-09-07
Combined ROS and control libraries images

domire8:
As I'm using the combination of ROS and control libraries quite often now, it would be nice to have those images prebuilt and pushed (same for modulo btw).

On top of the ROS workspace, control libraries (source, protocol, and python) is installed in the Dockerfiles. The branch of control libraries can be specified for the build. 

Opening this as draft PR because I'm waiting for [#190](https://github.com/epfl-lasa/control_libraries/pull/190) and [#198](https://github.com/epfl-lasa/control_libraries/pull/198) on control libraries. Also we could discuss the name of this image. Should this be called `modulo-base` or similar? Would be a bit more catchy but also less clear what it is :smile:


domire8:
This should be ready now, see [my fork](https://github.com/domire8/docker-images) for the result

ros2_control_libraries/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN git clone -b ${CL_BRANCH} --single-branch https://github.com/epfl-lasa/control-libraries.git
+
+# install source
+RUN sudo ./control-libraries/source/install.sh -y
+# install protocol
+RUN sudo ./control-libraries/protocol/install.sh -y
+# install python bindings
+RUN sudo pip3 install ./control-libraries/python
+# remove files except protocol
+RUN mv ./control-libraries/protocol ./protocol && rm -rf ./control-libraries/* && mv ./protocol ./control-libraries/protocol
```
eeberhard, line 22: I suggest to remove the entire directory. I don't know if the .proto files will be useful on their own to a ROS user, since they will benefit from the installed C++ and python integrations. If needed, the proto files can also be copied from the repository again later.

ros2_control_libraries/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+RUN sudo apt-get clean && sudo rm -rf /var/lib/apt/lists/*
```
eeberhard, line 24: Add the robotpkg key back as it is in modulo
`RUN sudo curl http://robotpkg.openrobots.org/packages/debian/robotpkg.key | sudo apt-key add -`

eeberhard (approved):
Looks good to go apart from the two comments

domire8:
Waiting for [that](https://github.com/epfl-lasa/control-libraries/pull/203) now

buschbapti (approved):

PR #15, opened 2021-10-02
Add autocompletion option

domire8:
There it is, the autocompletion function for our very own aica-docker hehe! 

The completion scripts need to be sourced in bashrc/zshrc so this is the only point I'm not totally sure how to do it. Maybe there should be an option for the install where the user can decide if he wants to install the auto completion or not...And then we need to make sure it works for at least bash/zsh on linux and mac...

### DIFF
--- a/scripts/install-aica-docker.sh
+++ b/scripts/install-aica-docker.sh
@@ -20,7 +20,7 @@ function get_shell_rc_path () {
 }
 
 function source_completion_script () {
-  echo "Writing the auto completion option to &1."
+  echo "Writing the auto completion option to $1."
   grep -v /src/aica-docker-completion.sh "$1" > tmpfile && mv tmpfile "$1"
   if [[ "$OSTYPE" != "darwin"* ]]; then
     echo "autoload bashcompinit; bashcompinit; source ${SCRIPT_DIR}/src/aica-docker-completion.sh" >> "$1"
@@ -47,7 +47,7 @@ sudo ln -s "${SCRIPT_DIR}/aica-docker.sh" ${SYMLINK}
 SHELL_RC_PATH=$(get_shell_rc_path)
 while true; do
   read -r -p "Do you with to install auto completion for aica-docker to '${SHELL_RC_PATH}'? [yes|no]
-[yes] >>>" yn
+>>>" yn
   case $yn in
     "" | yes) source_completion_script "${SHELL_RC_PATH}"; break;;
     no) exit;;
--- a/scripts/install-aica-docker.sh
+++ b/scripts/install-aica-docker.sh
@@ -20,6 +20,7 @@ function get_shell_rc_path () {
 }
 
 function source_completion_script () {
+  echo "Writing the auto completion option to &1."
   grep -v /src/aica-docker-completion.sh "$1" > tmpfile && mv tmpfile "$1"
   if [[ "$OSTYPE" != "darwin"* ]]; then
     echo "autoload bashcompinit; bashcompinit; source ${SCRIPT_DIR}/src/aica-docker-completion.sh" >> "$1"
@@ -48,7 +49,7 @@ while true; do
   read -r -p "Do you with to install auto completion for aica-docker to '${SHELL_RC_PATH}'? [yes|no]
 [yes] >>>" yn
   case $yn in
-    yes) source_completion_script "${SHELL_RC_PATH}"; break;;
+    "" | yes) source_completion_script "${SHELL_RC_PATH}"; break;;
     no) exit;;
     *) echo "Please answer 'yes' or 'no'.";;
   esac
--- a/scripts/install-aica-docker.sh
+++ b/scripts/install-aica-docker.sh
@@ -46,7 +46,7 @@ sudo ln -s "${SCRIPT_DIR}/aica-docker.sh" ${SYMLINK}
 
 SHELL_RC_PATH=$(get_shell_rc_path)
 while true; do
-  read -r -p "Do you with to install auto completion for aica-docker to '${SHELL_RC_PATH}'? [yes|no]
+  read -r -p "Do you with to install auto completion for aica-docker to '${SHELL_RC_PATH}'? [YES|no]
 >>>" yn
   case $yn in
     "" | yes) source_completion_script "${SHELL_RC_PATH}"; break;;
--- a/scripts/install-aica-docker.sh
+++ b/scripts/install-aica-docker.sh
@@ -23,9 +23,9 @@ function source_completion_script () {
   echo "Writing the auto completion option to $1."
   grep -v /src/aica-docker-completion.sh "$1" > tmpfile && mv tmpfile "$1"
   if [[ "$OSTYPE" != "darwin"* ]]; then
-    echo "autoload bashcompinit; bashcompinit; source ${SCRIPT_DIR}/src/aica-docker-completion.sh" >> "$1"
-  else
     echo "source ${SCRIPT_DIR}/src/aica-docker-completion.sh" >> "$1"
+  else
+    echo "autoload bashcompinit; bashcompinit; source ${SCRIPT_DIR}/src/aica-docker-completion.sh" >> "$1"
   fi
 }
 
--- a/scripts/install-aica-docker.sh
+++ b/scripts/install-aica-docker.sh
@@ -1,19 +1,26 @@
 #!/bin/bash
 
-function source_completion_script () {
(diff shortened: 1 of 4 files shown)


domire8:
Quick update on what I did in the last commit:
- The install checks for an existing symlink and updates the old one, if applicable.
- The install asks the user for permission (yes no input) to add a line to the bashrc/zshrc
- Due to the fact that (at least on my computer) the default shell of the sudo user is bash whilst mine is zshrc, you don't have to run the script with sudo permissions but rather the script has two sudo commands where the user will need to provide the password. This ensures that the auto completion script is written to the correct rc file.
- The uninstall only removes something if it exists.

All in all, much cleaner installing/uninstalling procedure to give the user clear messages and full control over the install.

scripts/install-aica-docker.sh
```diff
@@ -1,10 +1,50 @@
...
+  echo "Creating a symbolic link from ${SYMLINK} to ${SCRIPT_DIR}/aica-docker.sh;"
+fi
+sudo ln -s "${SCRIPT_DIR}/aica-docker.sh" ${SYMLINK}
 
-echo "Creating a symbolic link from /usr/local/bin/aica-docker to ${SCRIPT_DIR}"
-ln -s "${SCRIPT_DIR}" /usr/local/bin/aica-docker
+SHELL_RC_PATH=$(get_shell_rc_path)
+while true; do
+  read -r -p "Do you with to install auto completion for aica-docker to '${SHELL_RC_PATH}'? [yes|no]
+[yes] >>>" yn
```
buschbapti, line 44: This line is a bit weird as `yes` is not a default value so I don't know why it is there. Just put nothing before the read input. I would also suggest to make a default value, either yes or no let's say yes because most probably this is the choice wanted. So basically if I press enter then it is the yes choice, anything else is no.

buschbapti (commented):
This is a very good addition, just a few comments to make the script a bit better in terms of user input

scripts/install-aica-docker.sh
domire8: Okay but if `yes` if the default option, i should leave it there no?

buschbapti: Usually we put it capital in the `[Y|n]` on top

buschbapti: that indicates it is default or you actually write `default = yes`

scripts/src/aica-docker-completion.sh
```diff
@@ -0,0 +1,33 @@
...
+IMAGES=$(docker images --format "{{.Repository}}:{{.Tag}}")
+IMAGES=${IMAGES//"<none>:<none>"/}
+
+_aica_docker () {
+  local cur prev
+  COMPREPLY=()
+  cur="${COMP_WORDS[COMP_CWORD]}"
+  prev="${COMP_WORDS[COMP_CWORD-1]}"
+
+  if [[ ${COMP_CWORD} -eq 2 ]]; then
```
eeberhard, line 14: I haven't looked into the syntax of autocompletion too much yet, but if you regenerate the variables here on the completion call, is it needed to populate them in advance at the top of the file? If it is needed in both places, you could reduce code duplication by setting these variables in a sub function, though that's a bit awkward to parse and would increase the verbosity too. Mainly I want to check that you don't define it twice if you don't need to.

scripts/uninstall-aica-docker.sh
```diff
@@ -1,7 +1,31 @@
...
+  user=$(logname)
+  home_dir=$(eval echo "~${user}")
+  if [[ "${SHELL##*|}" == *"zsh"* ]]; then
+    echo "${home_dir}/.zshrc"
+  elif [[ "${SHELL##*|}" == *"bash"* ]]; then
+    if [[ "$OSTYPE" != "darwin"* ]]; then
+      echo "${home_dir}/.bashrc"
+    else
+      echo "${home_dir}/.bashrc"
+    fi
```
eeberhard, line 14: same comment here

scripts/install-aica-docker.sh
```diff
@@ -1,10 +1,56 @@
...
 
-echo "Creating a symbolic link from /usr/local/bin/aica-docker to ${SCRIPT_DIR}"
-ln -s "${SCRIPT_DIR}" /usr/local/bin/aica-docker
+SHELL_RC_PATH=$(get_shell_rc_path)
+while true; do
+  read -r -p "Do you with to install auto completion for aica-docker to '${SHELL_RC_PATH}'? [YES|no]
+>>>" yn
+  case $yn in
+    "") source_completion_script "${SHELL_RC_PATH}"; break;;
+    yes) source_completion_script "${SHELL_RC_PATH}"; break;;
```
eeberhard, line 52: you should be able to combine this as follows (but please test it first):
```suggestion
    "" | yes) source_completion_script "${SHELL_RC_PATH}"; break;;
```

scripts/install-aica-docker.sh
```diff
@@ -1,10 +1,56 @@
...
+}
+
+function source_completion_script () {
+  if grep -Rq "/src/aica-docker-completion.sh" "$1"; then
+    echo "Updating auto completion in $1"
+    grep -v /src/aica-docker-completion.sh "$1" > tmpfile && mv tmpfile "$1"
+  else
+    echo "Writing auto completion to $1"
+  fi
+  echo "source ${SCRIPT_DIR}/src/aica-docker-completion.sh" >> "$1"
```
eeberhard, line 28: For zshrc, you will need to add the following lines
```shell
autoload bashcompinit
bashcompinit
source ${SCRIPT_DIR}/src/aica-docker-completion.sh
```

I tested it like this and it works beautifully

eeberhard (approved):
Nice work! Install and uninstall script works great. 

The only thing that is needed in zshrc to support autocompletion are the lines:
```shell
autoload bashcompinit
bashcompinit
```

So, if you could install and remove those lines _if_ the shell is zsh, that would be perfect. It's very satisfying to use the scripts now with the autocompletion, thank you

domire8:
> Nice work! Install and uninstall script works great.
> 
> The only thing that is needed in zshrc to support autocompletion are the lines:
> 
> ```shell
> autoload bashcompinit
> bashcompinit
> ```
> 
> So, if you could install and remove those lines _if_ the shell is zsh, that would be perfect. It's very satisfying to use the scripts now with the autocompletion, thank you

That's a Darwin thing then ;) I have zsh and its not needed

scripts/install-aica-docker.sh
```diff
@@ -1,10 +1,56 @@
...
+  user=$(logname)
+  home_dir=$(eval echo "~${user}")
+  if [[ "${SHELL##*|}" == *"zsh"* ]]; then
+    echo "${home_dir}/.zshrc"
+  elif [[ "${SHELL##*|}" == *"bash"* ]]; then
+    if [[ "$OSTYPE" != "darwin"* ]]; then
+      echo "${home_dir}/.bashrc"
+    else
+      echo "${home_dir}/.bashrc"
+    fi
```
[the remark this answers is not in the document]
domire8: I have read somewhere that if you use bash on Mac, you'd have to write to bash_profile but since you dont use bash on Mac, I can't test that...

scripts/src/aica-docker-completion.sh
domire8: of course, thanks for the hint

scripts/install-aica-docker.sh
domire8: that does work :)

scripts/install-aica-docker.sh
```diff
@@ -1,10 +1,50 @@
...
-ln -s "${SCRIPT_DIR}" /usr/local/bin/aica-docker
+SHELL_RC_PATH=$(get_shell_rc_path)
+while true; do
+  read -r -p "Do you with to install auto completion for aica-docker to '${SHELL_RC_PATH}'? [yes|no]
+[yes] >>>" yn
+  case $yn in
+    yes) source_completion_script "${SHELL_RC_PATH}"; break;;
+    no) exit;;
+    *) echo "Please answer 'yes' or 'no'.";;
+  esac
```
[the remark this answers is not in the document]
eeberhard: @domire8 this is still a good point, you can put that echo in `source_completion_script` function

eeberhard (approved):

buschbapti (approved):
All good now. Thanks

buschbapti (approved):

eeberhard (approved):

PR #16, opened 2021-10-04
Update workflows with repository owner variable

domire8:
As I've been working on the CI a bit to check for updated repositories, I've come across some points that can be improved. As I don't want to put everything in one PR, this is the first bit of it.
Instead of hardcoding `aica-technology` to the image names everywhere, we can just use `${{ github.repository_owner }}`, a much cleaner solution.

Tested on my fork.

### DIFF
--- a/.github/actions/build-push/action.yml
+++ b/.github/actions/build-push/action.yml
@@ -4,7 +4,7 @@ inputs:
   image:
     description: 'The desired image'
     required: false
-    default: 'aica-technology/ros2-ws:foxy'
+    default: 'ros2-ws:foxy'
   cl_branch:
     description: 'The branch of control libraries'
     required: false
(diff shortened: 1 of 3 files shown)


buschbapti (approved):
Well that seems like a very neat improvement. I like that.

eeberhard (approved):
Great, seems like a nice improvement. Careful that this syntax is only used for properly formatted repository names (e.g. the prior `control_libraries` might have been an issue). Here it's good for sure

domire8:
> Great, seems like a nice improvement. Careful that this syntax is only used for properly formatted repository names (e.g. the prior `control_libraries` might have been an issue). Here it's good for sure

Remind me, is that because of our own conventions, or because it would actually not be supported?

PR #17, opened 2021-10-05
Add scheduled workflow to check for updates control libraries

domire8:
Here is a scheduled workflow that check for the latest commit hash of the develop branch of  control libraries and rebuilds the images if necessary. This workflow requires a branch called `ci` that is equivalent to main except that it has a file called `control-libraries-hash` containing the hash of the commit on control libraries.

Whats going on step by step:
1. The workflow checks out the ci branch of this repository, then checks if there is a more recent hash on the develop branch of control libraries.
2. If this is the case, the main branch of this repository is checked out, and the `ros2-control-libraries:foxy` image is rebuilt.
3. If step 2 is successful, the ci branch of this repository is checked out again, rebased onto main and the new hash is written to the file and commited as github actions user.

The workflow is running twice a day, once at night and once during the day, I think that should be enough but I'm open to discuss this.

Note that only the foxy image is rebuilt here, I don't think we need to include noetic for now because I'm the only one that uses it from time to time anyway, so I'll do it manaully.

Before merging, I just need to create the ci branch as described above.

See my fork for confirmation that it works: https://github.com/domire8/docker-images


eeberhard (approved):
Excellent, this implementation is exactly as we discussed. I think it's great to have it all in one repo here, so that if we add more images and triggers for other upstreams (e.g. modulo, control libraries) it's all centralised.

Just a couple of minor comments but overall great job

eeberhard (approved):

PR #18, opened 2021-10-05
Robustify uninstall

eeberhard:
* The `test -f ${FILE}` line should return true for files and
symlinks, but in case the aica-docker file in the bin is
corrupted for whatever reason it might as well use a silent
rm -f even if the check returns false

This hack should fix @buschbapti's uninstall issue (running uninstall did not remove the symlink from bin). It wait until after #15 

@domire8

### DIFF
--- a/scripts/uninstall-aica-docker.sh
+++ b/scripts/uninstall-aica-docker.sh
@@ -3,5 +3,6 @@
 SYMLINK=/usr/local/bin/aica-docker
 if test -f "$SYMLINK"; then
   echo "Removing symbolic link ${SYMLINK}"
-  rm "${SYMLINK}"
 fi
+
+rm -f "${SYMLINK}"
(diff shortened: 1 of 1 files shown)


domire8 (approved):
All good, let's do this one first

scripts/uninstall-aica-docker.sh
```diff
@@ -3,5 +3,6 @@
 SYMLINK=/usr/local/bin/aica-docker
 if test -f "$SYMLINK"; then
```
buschbapti, line 4: I feel this was the line that was not working. As it did not print the echo. So I would prefer that we remove the if and print the echo line even if the file is not here. Otherwise, here, it would actually delete the file on my side but do not print the echo which I think is worse.

buschbapti: Confirmed that removing the `if` statement solved my issue

buschbapti (approved):

PR #20, opened 2021-10-07
Build modulo and ros2control

eeberhard:
AICA private repositories heavily use modulo and ros2 control dependencies. It makes sense to host these base images here in this repository, to manage and control the versions across the stack, and to prevent local rebuilding whenever modulo is updated. 

The reason we build modulo core here and don't pull from the modulo CI is that the modulo base image may be out of date (the modulo CI does not check for control libraries updates).

I also left `ros2-modulo` and `ros2-modulo-control` separate so it's easier to mix and match branches / versions if we need it.

Your careful review and comments are welcome

@buschbapti @domire8


eeberhard:
I agree it's a bit hard to test, and really I should fork it... but it's not "production critical" if the CI on main fails either so I think we should just see what happens :)

PR #24, opened 2021-11-10
Add option to build ROS2 modulo control image with galactic

domire8:
Most of this PR are some minor improvements from @eeberhard that I propagated to the ROS1 images too. The main change is in the Dockerfile of ros2-modulo-control. Due to the fact that the CI doesnt have the BuildKit, its a bit trickier to enable two different installation procedures for foxy and galactic. But the way I did it works and if we start using galactic we should have it in CI as well.

That being said, I should maybe add galactic to the default CI too, so it builds automatically alongside noetic and foxy. What do you think?


ros2_modulo_control/build.sh
```diff
@@ -33,4 +33,4 @@ fi
 BUILD_FLAGS+=(--build-arg ROS_VERSION="${ROS_VERSION}")
 BUILD_FLAGS+=(-t "${IMAGE_NAME}:${ROS_VERSION}")
 
-DOCKER_BUILDKIT=1 docker build "${BUILD_FLAGS[@]}" .
+DOCKER_BUILDKIT=0 docker build "${BUILD_FLAGS[@]}" .
```
buschbapti, line 36: this should be kept

buschbapti (approved):
Very clean thank and useful thank you

ros2_modulo_control/build.sh
domire8:
```suggestion
DOCKER_BUILDKIT=1 docker build "${BUILD_FLAGS[@]}" .
```

eeberhard (approved):
Looks great, thanks for following up on my test branch with those improvements.

Can I ask what you did with the CI for `ros2_modulo_control/Dockerfile` - you mentioned it wouldn't work without buildkit to have the ARG-based layering, but in the end it looks to be the same structure that I had originally. I assume it's because the layering is not actually dependent on buildkit (which is instead needed for things like better layer caching and stopping the build on a specific target layer).

domire8:
> Looks great, thanks for following up on my test branch with those improvements.
> 
> Can I ask what you did with the CI for `ros2_modulo_control/Dockerfile` - you mentioned it wouldn't work without buildkit to have the ARG-based layering, but in the end it looks to be the same structure that I had originally. I assume it's because the layering is not actually dependent on buildkit (which is instead needed for things like better layer caching and stopping the build on a specific target layer).

It is very similar indeed. The main difference is that the two first steps (the one for foxy and the one for galactic) are always built because you cannot skip layers without the buildkit. So for those two steps, the ROS version is actually hardcoded and only in the last layer, the ARG ROS_VERSION is defined and will just continue from the correct step. You initially had the ARG ROS_VERSION for all layers which then doesnt work if you want to install ros-galactic stuff with apt-get.

buschbapti:
Just added the option to pass the `ROS_DOMAIN_ID` environment variable. Since moving to `galactic` network discovery is now back correctly so it was annoying to see the other nodes of people in the lab...

domire8:
Tested and it works as expected, thanks @buschbapti. Do you agree to this too @eeberhard ?

scripts/src/server.sh
```diff
@@ -50,6 +51,9 @@ Options:
                            For the list of gpu_options parameters see:
     >>> https://docs.docker.com/config/containers/resource_constraints/
 
+  --ros-domain-id <id>     Set the ROS_DOMAIN_ID environment variable
+                           to avoid conflicts when doing network discovery
```
eeberhard, line 55:
```suggestion
                           to avoid conflicts when doing network discovery.
```

eeberhard (approved):
Domain ID is a good addition, and as discussed it makes sense to set a consistent default when running containers through our script ecosystem.

For reference, the ID 14 was chosen as the sum of indices for the word `aica -> 1 + 9 + 3 + 1`.

Happy with this to be merged provided it has been tested successfully!

domire8:
Tested, merging now

PR #25, opened 2021-11-12
Fix permission issues

domire8:
Sorry guys, I didn't rebuild the images locally before pushing. There was a small permission issue that is now fixed.


eeberhard (approved):
It's all good, that's what the CI is for! Thanks for touching up the docs too

PR #27, opened 2022-01-12
Remove foxy images from CI builds

domire8:
I did two things here, hence two commits.
- change all the default tags and args in the build scripts and Dockerfile to galactic
- remove the foxy and noetic images from the automatic CI. Note that its still possible to build those through the manual build action so the only thing we loose is the *automatic* build, we can still do it manually

### DIFF
--- a/.github/actions/build-push/action.yml
+++ b/.github/actions/build-push/action.yml
@@ -4,7 +4,7 @@ inputs:
   image:
     description: 'The desired image'
     required: false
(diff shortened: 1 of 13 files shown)


eeberhard (approved):
Perfect! I agree it's nice to reduce the amount of computation steps in the automatic build when we don't need them, and we always have manually configured and triggered build available when needed.

PR #28, opened 2022-01-31
Python installation from source

eeberhard:
* Install a specific python version from source

This adds python installation from source, so that we can choose precisely what version of python we want as the system default. It requires a reasonable amount of time to build, and some extra build tools. Thankfully, this base image very rarely has to be rebuilt (usually once a month or so), so the extra build time is not a big problem. And, I think a lot of the build tools already existed in the base image, but it would take some effort to check each one manually, so for now we just list them all in the initial `apt-get install`. Finally, perhaps some of the build tools could be removed from the image at the end to reduce file size, but that would also be best as a separate refactor.

When it was built locally, I had `python3 --version >> 3.10.2` as expected

### DIFF
--- a/ros2_ws/Dockerfile
+++ b/ros2_ws/Dockerfile
@@ -5,23 +5,35 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && apt-get install -y \
   autoconf \
   automake \
+  build-essential \
   curl \
   featherpad \
   gdb \
   git \
   iputils-ping \
   libboost-all-dev \
+  libbz2-dev \
+  libncurses5-dev \
+  libncursesw5-dev \
+  libreadline-dev \
+  libsqlite3-dev \
+  libssl-dev \
   libtool \
+  llvm \
+  make \
   mesa-utils \
   nano \
-  python3-pip \
   ros-${ROS_DISTRO}-xacro \
   ros-${ROS_DISTRO}-robot-state-publisher \
   ros-${ROS_DISTRO}-rviz2 \
   rsync \
   software-properties-common \
   ssh \
+  tk-dev \
   unzip \
+  wget \
+  xz-utils \
+  zlib1g-dev \
   && rm -rf /var/lib/apt/lists/*
 
 RUN echo "Set disable_coredump false" >> /etc/sudo.conf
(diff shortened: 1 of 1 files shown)


ros2_ws/Dockerfile
```diff
@@ -5,23 +5,35 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && apt-get install -y \
   autoconf \
   automake \
+  build-essential \
```
domire8, line 8: Why do we need so many additional libraries?

domire8 (approved):
Looks good to me

ros2_ws/Dockerfile
eeberhard: This is the recommended set of libraries for installing python from source, which includes the compiler optimisations, as well as system libraries that the python built-ins use at runtime (like libssl for example). For `build-essential` and some others I am almost certain they are already there in the base image.

We could also install from package `apt-get install python3.10` but from what I saw this anyway will download a lot of these libraries required for runtime and testing, so it doesn't save that much space or time in the grand scheme of things. Also the specific versions of python available on apt depends on the apt repository, not all are always easily available.

buschbapti (approved):
Thanks for that seems good to me and a good way to go to keep track of our dependencies as well.

ros2_ws/Dockerfile
```diff
@@ -35,7 +47,19 @@ RUN ( \
   && mkdir /run/sshd
 
 
-FROM base-dependencies as base-workspace
+FROM base-dependencies as python3-install
+ARG PYTHON_VER=3.10.2
+WORKDIR /tmp
+
+RUN wget https://www.python.org/ftp/python/${PYTHON_VER}/Python-${PYTHON_VER}.tar.xz
+RUN sudo tar -xf Python-${PYTHON_VER}.tar.xz
```
domire8, line 55: I missed this before but why do you do `sudo` here? you are already root here

eeberhard: ah it's true, good point. I was used to having derived docker images start as USER ros2, but of course this is the base image. I can address that as part of the fix in #29

domire8:
Is that something you would like to try again? @eeberhard

eeberhard:
> Is that something you would like to try again? @eeberhard

I don't really see it as a priority. The main issue was ROS2 compatibility with python 3.10, and I don't know if that has improved since Humble. In my opinion we don't have to act on this for now.

PR #29, opened 2022-02-01
Install catkin-pkg

eeberhard:
* Explicitly install the catkin packages for the python3 interpreter.

When installing python3 from source as the default system interpreter, the catkin-pkg module could not be found by ament_cmake, which is invoked during colcon build. This naturally caused all downstream ROS2 workspace builds to fail.

Now colcon build succeeds (tested locally to build modulo image with most recent ghcr.io/aica-technology/ros2-control-libraries:galactic)


domire8 (approved):
Thanks for the sudo fix, that was what I was going to propose. After #30, I'm happy if this goes in, I can confirm all the images build now

domire8 (changes requested):
Actually, cancel that, The ros2_modulo_control image doesn't build

eeberhard:
very well, and so we find our way down to the next issue. Indeed I only tested the modulo stage. What is the error output you got from the subsequent ros2 control build attempt?

ros2_ws/Dockerfile
```diff
@@ -52,11 +52,13 @@ ARG PYTHON_VER=3.10.2
...
+RUN tar -xf Python-${PYTHON_VER}.tar.xz
 WORKDIR Python-${PYTHON_VER}
-RUN sudo ./configure --enable-optimizations --with-ensurepip=install \
-  && sudo make -j8 \
-  && sudo make install
+RUN ./configure --enable-optimizations --with-ensurepip=install \
+  && make -j8 \
+  && make install
+
+RUN python3 -m pip install catkin_pkg
```
domire8, line 61:
```suggestion
RUN ./configure --enable-optimizations --with-ensurepip=install --enable-shared \
  && make -j8 \
  && make install
RUN ldconfig

RUN python3 -m pip install catkin_pkg empy numpy lark
```

domire8 (commented):
With those changes below, even ros2 modulo control builds now, you just get warnings like this for several packages:
```
#15 141.3 --- stderr: controller_manager
#15 141.3 <string>:1: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
#15 141.3 <string>:1: DeprecationWarning: The distutils.sysconfig module is deprecated, use sysconfig instead
#15 141.3 ---
``

eeberhard:
ROS2 Galactic expects Python 3.8, and it turns out that it is harder to upgrade than I expected. With 3.9.X, the backend builds (all components and state engine). However, it still fails at runtime. Specifically, it cannot resolve the `rosidl_typesupport_c` runtime library.

With what I've tried so far, `colcon build` will still make a mess and install some things to `/path/to/workspace/install/package_name/lib/python3.8/site-packages` and `/path/to/workspace/install/other_package_name/lib/python3.9/site-packages` on the PYTHONPATH. 

This is probably fixable but needs to be put on hold until it is more stable. Until then we must revert the original #28 or risk our downstream targets continuing to fail.

PR #32, opened 2022-02-11
Reduce installation size

eeberhard:
* Configure apt to not install recommended or suggested packages by 
default

* Configure pip to disable cache

This leads to a small but useful reduction in the image size of ~200MB. I tested the build pipeline locally and made sure the downstream targets (dynamic state engine) still builds and works at run time.

Also when I closed #31 to revert #28, I didn't realise it also reverted #30 - so the local base changes are re-included here.

Image sizes before and after:
- ros2-ws: 1.72 GB | 1.52 GB
- ros2-control_libraries: 2.65 GB | 2.47 GB
- ros2-modulo: 2.77 GB | 2.58 GB
- ros2-modulo-control: 3.29 GB | 3.14 GB

### DIFF
--- a/ros2_control_libraries/build.sh
+++ b/ros2_control_libraries/build.sh
@@ -2,7 +2,7 @@
 
 IMAGE_NAME=aica-technology/ros2-control-libraries
 
-LOCAL_BASE_IMAGE=0
+LOCAL_BASE_IMAGE=false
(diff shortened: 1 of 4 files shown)


domire8 (approved):
Yep looks good!
Can we delete the fix/revert branch?

buschbapti (approved):
That's some nice optimization there.

PR #33, opened 2022-02-14
Install wget

domire8:
I'm using `wget` in the network interfaces library script, to download the cppzmq bindings. Would be better if we can install this here, it just increases the size of the image from 1.52 to 1.53GB


buschbapti (approved):

domire8:
This was probably due to the -no-install-recommends change FYI

eeberhard (approved):

PR #34, opened 2022-02-18
Tag development images

eeberhard:
I want us to have a stable "main" release of our docker images alongside "development" images (which use development branches of control libraries and modulo). For this reason I propose to add more tagging logic to our build process, and by default generate a `:galactic` and `:galactic-devel` tag of our composite images.

For now `-devel` only signifies the branches of control libraries and modulo, but as you can see from the ros2-modulo-control docker file, the `devel` stage would also let us safely try newer version of ros2 control without breaking our main image.

It's more images, so more compute and hosting - but the full set of images (main and development) are only rebuilt on push to main or manual dispatch, while the check-upstream will handle the `-devel` side. We just have to manually dispatch when there is a new main release.

I forked this and tested it successfully here (with minor modifications so the base images pull from `eeberhard` instead of `aica-technology`):
https://github.com/eeberhard/docker-images-test/actions/runs/1864987461

As you can see, the packages are generated with the `galactic` and `galactic-devel` tags for [control libraries](https://github.com/eeberhard/docker-images-test/pkgs/container/docker-images-test%2Fros2-control-libraries), [modulo](https://github.com/eeberhard/docker-images-test/pkgs/container/docker-images-test%2Fros2-modulo) and [modulo-control](https://github.com/eeberhard/docker-images-test/pkgs/container/docker-images-test%2Fros2-modulo-control).

--- 

* Refactor Dockerfiles and build scripts to use BASE_TAG notation instead of ROS_VERSION for build args, to allow more diverse range of base tags when pulling from images

* Refactor build push-action inputs to increase configurability, supplying workspace and tags separately instead of compositely. Add inputs for ros version and base and output tags

* Add step to build-push action to parse workspace, tags and image name into environment variables for later steps

* Update ros2-modulo-control dockerfile to support a galactic-devel build stage

* Add development stages to build-push workflow to use development branches of control libraries and modulo, tagging them with the respecitve galactic-devel tags. Change default galactic tag images to use main branches of control libraries and modulo

* Update check-upstream and manual-dispatch workflows with new action input syntax

### DIFF
--- a/.github/actions/build-push/action.yml
+++ b/.github/actions/build-push/action.yml
@@ -1,10 +1,14 @@
 name: 'Build and Push'
 description: 'Build the Docker image and push to GitHub Container Registry'
 inputs:
-  image:
-    description: 'The desired image'
+  workspace:
+    description: 'The desired workspace'
+    required: true
+    default: 'ros2-ws'
+  ros_version:
+    description: 'The ROS distribution to use'
     required: false
-    default: 'ros2-ws:galactic'
+    default: 'galactic'
   cl_branch:
     description: 'The branch of control libraries'
     required: false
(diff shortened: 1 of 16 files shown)


buschbapti (approved):

domire8:
Looks good to me, it's tested on your fork so I trust it :) always hard to check the CI otherwise

domire8 (approved):

PR #35, opened 2022-02-22
Remove foxy and extend build scripts

domire8:
So @eeberhard and me decided to remove foxy from the `ros2-modulo-control` image because 
- this Dockerfile is kind of annyoing with the different stages and we will not switch back to foxy anyway
- the new improved manual dispatch does not work with other tags than foxy, galactic and galactic-devel, which is kind of sad as well

Therefore, I decided to adapt the build scripts and their arguments according to the new workflow dispatch options - i.e. that you can provide a base tag and an output tag (except for the workspace image where the base tag as well as the output tag are the ROS version). I also took the chance to add verbose and rebuild options to the build scripts

### DIFF
--- a/ros2_control_libraries/build.sh
+++ b/ros2_control_libraries/build.sh
@@ -5,7 +5,7 @@ IMAGE_NAME=aica-technology/ros2-control-libraries
 LOCAL_BASE_IMAGE=false
 BASE_IMAGE=ghcr.io/aica-technology/ros2-ws
 BASE_TAG=galactic
-EXPORT_TAG=galactic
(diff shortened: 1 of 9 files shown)


eeberhard (approved):
Good change, though it's a shame the build scripts have so much code duplication. It becomes apparent in a PR like this when you have to make the same changes across all the scripts!

If you have tested the pipeline locally, then the change to the dockerfile should be good in the CI too.

buschbapti (approved):
Nice thanks. I agree with Enrico here but not sure how we can change that.

PR #36, opened 2022-03-10
Write sha256 hashes of images to file

domire8:
With these changes, the CI writes the sha256 hashes of generated images to a file in the workspace folder. For example, after building `ros2-modulo:galactic-devel` there will be a file on the `ci` branch under `ros2_modulo/galactic-devel-hash` containing the sha256 hash of the pushed image. This allows downstream repos to check for updated images.

For the manual workflow dispatch, where we can generate any kind of tags, there is the option to disable writing the hash to file.

See https://github.com/domire8/docker-images/tree/ci

### DIFF
--- a/.github/actions/build-push/action.yml
+++ b/.github/actions/build-push/action.yml
@@ -29,6 +29,10 @@ inputs:
     description: 'Write image SHA256 hash to file'
     required: false
     default: 'true'
+  ci_branch:
+    description: 'The name of the CI branch'
+    required: false
+    default: 'ci'
   secret:
     description: 'GitHub Container Registry secret'
     required: true
--- a/.github/actions/build-push/action.yml
+++ b/.github/actions/build-push/action.yml
@@ -91,7 +91,7 @@ runs:
       run: |
         git config user.name github-actions
         git config user.email 41898282+github-actions[bot]@users.noreply.github.com
-        git fetch && git checkout ci && git pull && git rebase origin/main
+        git fetch origin ${{ inputs.ci_branch }} && git checkout ${{ inputs.ci_branch }} && git rebase origin/main
         SHA=$(docker images --no-trunc --quiet ${IMAGE_NAME})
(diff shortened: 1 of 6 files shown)


.github/actions/build-push/action.yml
```diff
@@ -79,3 +85,15 @@ runs:
...
+    - name: Write new image hash to file
+      if: ${{ inputs.write_image_hash == 'true' }}
+      run: |
+        git config user.name github-actions
+        git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+        git fetch && git checkout ci && git pull && git rebase origin/main
+        SHA=$(docker images --no-trunc --quiet ${IMAGE_NAME})
+        echo ${SHA} > ./${{ env.WORKSPACE_PATH }}/${{ env.OUTPUT_TAG }}-hash
+        git add ./${{ env.WORKSPACE_PATH }}/${{ env.OUTPUT_TAG }}-hash
+        git commit -m "Update ${{ env.IMAGE_NAME }} image hash" && git push -f
```
eeberhard, line 98: In your fork, the CI branch has all SHA files in one commit with the message `Update ros2-modulo:galactic-devel image hash`. From seeing this build-push action I would have expected one commit per image (e.g. one commit per invocation of the build-push action from the workflow). Am I missing something?

domire8: something is wrong indeed, thanks for pointing out...i have to investigate

eeberhard (approved):
Excellent work!

buschbapti (approved):

domire8:
okay this should be it, i also fixed the error we had from the previous PR, tested here https://github.com/domire8/docker-images/actions

eeberhard (approved):
Thanks for testing again. I don't know why `git fetch origin ci:ci` wasn't working - that's how I regularly do background "fetch pulls". For example if I'm on a feature branch and want to update my `develop` branch in the background, I'll do `git fetch origin develop:develop` and subsequently can just `git rebase develop` with it being up to date.

Anyway your fork looks good, all commit messages and generated hashes on the CI branch seem correct.

PR #38, opened 2022-03-21
Improve CI

domire8:
Another fix I want to merge before #36, review by commit:

Recreate CI branch on push on main: Each time there is a push on main, the CI branch is recreated to delete the possibly long commit history from it and to avoid rebasing conflicts.

Add action to write hash to file: I created a new action that logs in as github user and writes a desired hash to file, mostly to reduce code duplication

Fix if conditions in upstream workflow: I had to move around some if conditions in the upstream checking workflow because if the control libraries rebuild doesn't run at all, the modulo rebuild will never run either, which is not desired because only modulo might change.

### DIFF
--- a/.github/actions/write-hash/action.yml
+++ b/.github/actions/write-hash/action.yml
@@ -21,7 +21,7 @@ runs:
       run: |
         git config user.name github-actions
         git config user.email 41898282+github-actions[bot]@users.noreply.github.com
-        git fetch && git checkout ${{ inputs.ci_branch }} && git pull && git rebase origin/main
+        git fetch origin ${{ inputs.ci_branch }}:${{ inputs.ci_branch }} && git checkout ${{ inputs.ci_branch }} && git rebase origin/main
         echo ${{ inputs.hash }} > ${{ inputs.file }}
         git add ${{ inputs.file }}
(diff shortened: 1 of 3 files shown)


domire8:
all tested here (see last 4 successful workflows): https://github.com/domire8/docker-images-2/actions

.github/actions/write-hash/action.yml
```diff
@@ -0,0 +1,28 @@
...
+    default: 'ci'
+
+runs:
+  using: "composite"
+  steps:
+    - name: Write hash to file
+      run: |
+        git config user.name github-actions
+        git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+        git fetch && git checkout ${{ inputs.ci_branch }} && git pull && git rebase origin/main
```
eeberhard, line 24: You can avoid a global fetch and subsequent pull by fetch-pulling the `ci` branch specifically.
```suggestion
        git fetch origin ${{ inputs.ci_branch }}:${{ inputs.ci_branch }} && git checkout ${{ inputs.ci_branch }} && git rebase origin/main
```

eeberhard (approved):

PR #39, opened 2022-04-07
Add workflow to periodically check for new commits on main branches

domire8:
I think we've all become too lazy and preoccupied with other things to always manually trigger a rebuild of our `galactic` images if modulo or control libraries have a new commit on main. I'm adding therefore another check upstream workflow that checks twice a week for new commits on main branches and rebuilds the images if necessary

### DIFF
--- a/.github/workflows/check-upstream-devel.yml
+++ b/.github/workflows/check-upstream-devel.yml
@@ -1,0 +1,124 @@
+name: Check for updated upstream libraries (develop branches)
+
+# Run scheduled workflow
+on:
+  schedule:
+   - cron: '20 3,13 * * MON-FRI'
+  workflow_dispatch:
+
+env:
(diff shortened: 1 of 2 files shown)


domire8:
I just thought about it, it's a lot of code duplication but I cannot create an action with arguments from it because I cannot use local actions from other local actions. We'd have to publish our basic actions like `write-hash` or `build-push`  to the github marketplace (which could be nice because we could reuse them in other repos, but also not very necessary at this point)

eeberhard (approved):
There's a bit of duplication but overall it's quite clean. It seems correct as far as I can tell, to refer to `-develop-hash` and `galactic-devel` in one and `-main-hash` and `galactic` in the other.

I had completely forgot that releases on main would not trigger these changes, meaning we had much older version of control libraries / modulo in the non-devel downstream images than I realised.

PR #40, opened 2022-04-07
Update build-push workflow to write both develop and main hashes

domire8:
Didn't think it through, there are other workflows writing those hashes to file


eeberhard (approved):
Ah of course, my mistake as well for not spotting that.

PR #41, opened 2022-04-14
Set cron correctly for check upstream main

domire8:
Just a small fix to a scheduled action. I saw that it ran 5 times instead of just once so I realized that I need to set a minute of the hour, otherwise it will run as many times as it can during the desired hour.


eeberhard (approved):
Interesting, good catch!

PR #45, opened 2022-05-11
Enable colorized ROS logging

domire8:
Will open a PR on state engine as well, so it works with python components too

@eeberhard


eeberhard (approved):
🔥

PR #46, opened 2022-06-23
Check for both default ed25519 and RSA SSH files and exit if not found

buschbapti:
This PR adds the checking of the existence of the provided SSH key file in the `server.sh` scripts as well as consider both `ed25519` and `rsa` files as default.

### DIFF
--- a/scripts/src/server.sh
+++ b/scripts/src/server.sh
@@ -133,11 +133,14 @@ if [ -z "$CONTAINER_NAME" ]; then
 fi
 
 if [[ ! -f "${SSH_KEY_FILE}" ]]; then
+  if [[ "${SSH_KEY_FILE}" != "${DEFAULT_SSH_ED25519}" ]]; then
+    echo "Provided SSH key file ${SSH_KEY_FILE} does not exist."
+    exit 1
+  fi
   if [[ ! -f "${DEFAULT_SSH_RSA}" ]]; then
     echo "SSH not set up! Configure SSH on your system."
     exit 1
   fi
-  echo "Provided SSH key file ${SSH_KEY_FILE} does not exist. Using default ${DEFAULT_SSH_RSA}"
   SSH_KEY_FILE="${DEFAULT_SSH_RSA}"
 fi
 PUBLIC_KEY=$(cat "${SSH_KEY_FILE}")
--- a/scripts/src/server.sh
+++ b/scripts/src/server.sh
@@ -1,7 +1,9 @@
 #!/bin/bash
 
 SSH_PORT=3333
-SSH_KEY_FILE="$HOME/.ssh/id_rsa.pub"
+DEFAULT_SSH_RSA="$HOME/.ssh/id_rsa.pub"
+DEFAULT_SSH_ED25519="$HOME/.ssh/id_ed25519.pub"
+SSH_KEY_FILE="${DEFAULT_SSH_ED25519}"
 IMAGE_NAME=""
 USERNAME=root
 GPUS=""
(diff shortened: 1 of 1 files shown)


scripts/src/server.sh
```diff
@@ -130,6 +132,14 @@ if [ -z "$CONTAINER_NAME" ]; then
   CONTAINER_NAME="${CONTAINER_NAME/:/-}-ssh"
 fi
 
+if [[ ! -f "${SSH_KEY_FILE}" ]]; then
+  if [[ ! -f "${DEFAULT_SSH_RSA}" ]]; then
+    echo "SSH not set up! Configure SSH on your system."
+    exit 1
+  fi
+  echo "Provided SSH key file ${SSH_KEY_FILE} does not exist."
+  exit 1
```
domire8, line 141: Wait why do you exit here? Shouldn't you do `${SSH_KEY_FILE} = ${DEFAULT_SSH_RSA}` instead

buschbapti: right yes indeed good catch

scripts/src/server.sh
```diff
@@ -130,6 +132,14 @@ if [ -z "$CONTAINER_NAME" ]; then
   CONTAINER_NAME="${CONTAINER_NAME/:/-}-ssh"
 fi
 
+if [[ ! -f "${SSH_KEY_FILE}" ]]; then
+  if [[ ! -f "${DEFAULT_SSH_RSA}" ]]; then
+    echo "SSH not set up! Configure SSH on your system."
+    exit 1
+  fi
+  echo "Provided SSH key file ${SSH_KEY_FILE} does not exist. Using default ${DEFAULT_SSH_RSA}"
```
eeberhard, line 140: Because you set the EDRSA file as the default `SSH_KEY_FILE` above, in most cases we will see this printed line, which is kind of annoying. In particular, if I don't supply a `-k` argument, I would be confused by this message saying something about `id_ed25519.pub`. There's no perfect solution here because someone might have multiple key pairs for different reasons. Preferably you just print out one line at the end of this whole if block that says "using SSH key file ${}". 

I would say the RSA should be default, and if it doesn't exist then choose EDRSA if it exists. If both exist, it will be RSA by default.

eeberhard (commented):
The change is justified based on our conversation. Just a minor comment

scripts/src/server.sh
```diff
@@ -130,6 +132,17 @@ if [ -z "$CONTAINER_NAME" ]; then
...
+if [[ ! -f "${SSH_KEY_FILE}" ]]; then
+  if [[ "${SSH_KEY_FILE}" != "${DEFAULT_SSH_RSA}" ]]; then
+    echo "Provided SSH key file ${SSH_KEY_FILE} does not exist."
+    exit 1
+  fi
+  if [[ ! -f "${DEFAULT_SSH_ED25519}" ]]; then
+    echo "SSH not set up! Configure SSH on your system."
+    exit 1
+  fi
+  SSH_KEY_FILE="${DEFAULT_SSH_RSA}"
```
eeberhard, line 144:
```suggestion
  else
    SSH_KEY_FILE="${DEFAULT_SSH_ED25519}"
  fi
```

buschbapti:
Should be good now.

scripts/src/server.sh
eeberhard: ^ with the else statement I think it's a bit more readable

eeberhard (approved):
Good to go if it works on your system and fixes your issue 👍

scripts/src/server.sh
buschbapti: do you prefer a `if ... else if ... else` then? I am not a huge fan of `else` in case of `exit` conditions but the `if ... else` here would only make sense with the `else if` in the middle I would say.

domire8 (approved):

PR #47, opened 2022-06-24
Replace ros_version with base_tag to fix action

domire8:
As we found out, the workflows were providing a `BASE_TAG` to the docker build stage but for the `ros(2)_ws` the Dockerfiles needed a `ROS_VERSION`. For consistency, I changed everything to `BASE_TAG` and nothing needs the ros version anymore.


buschbapti (approved):

eeberhard (approved):
Consistency is good, thank you!

PR #49, opened 2022-08-10
Add openrobots to CMake prefix path

domire8:
A small change to avoid having to append that path in every ROS package CMakeLists that is using robot model.


eeberhard (approved):
LGTM 👍

PR #50, opened 2022-08-15
Colcon test alias and persistent environment

eeberhard:
* Add colcon test helper with similar structure to the build helper that allows selecting specific packages for testing and automatically displays the test result

* Make environment variables persist in SSH instances by stashing environment variables into a file in the sshd entrypoint script

* Change into the ROS2_WORKSPACE directory temporarily when invoking the colcon build or test helpers

* Reorganize colcon defaults and helpers into a specific config/colcon subdirectory

* Simplify Dockerfile by invoking colcon utility setup script which handles the sourcing and aliases

--- 

TL;DR: you can now do `cb package_foo` or `ct package_foo` from anywhere in the container, whether running interactively or logging in via SSH, and it will cd to the `ROS2_WORKSPACE` and build / test the listed package(s) automatically before returning you to your current directory. You can also `ccd` to go directly to `ROS2_WORKSPACE`, or `ccd package_foo` to go to the source folder of the package (e.g. `/home/ros2/ros2_ws/src/my_packages/package_foo`).


eeberhard:
This was tested with a local build of `ros2_modulo` image and is working as intended. I had a small bug with passing the environment variable to the bash script but that is resolved now. An additional benefit of persisting the environment variables in the sshd entrypoint is that it also increases the consistency between `aica-docker interactive` and `aica-docker server`

domire8:
I was hoping this would solve my problem too but it doesn't.. When you start the server for modulo-controllers and try to load the CMakeLists in CLion, it fails because it cannot find pinocchio library. Once you added `LIST(APPEND CMAKE_PREFIX_LIST /opt/openrobos)` it works fine. Can you maybe try to replicate this and check why that happens?

eeberhard:
Thanks @domire8 for the catch. I retested it with modulo controllers and found the issue. I needed to prepend the source directive to `bashrc`, not append it, so that it happens first. In non-interactive shells (e.g. headless connection from CLion), `bashrc` breaks early and does not reach the end of the file. This is the same situation as the environment sourcing for the colcon workspace - it has to happen at the top of the file. 

Now the develop branch of modulo controllers builds with the new local image:
```console
-- Found modulo_core: 2.0.3 (/home/ros2/ros2_ws/install/modulo_core/share/modulo_core/cmake)
-- pinocchio FOUND. pinocchio at /opt/openrobots/lib/libpinocchio.so
```

domire8 (approved):
Cool! In that case, good to go

buschbapti (approved):

PR #51, opened 2022-08-23
Correctly escape ROS2_WORKSPACE variable in bashrc

eeberhard:
* Use single quotes to escape the variable written to bashrc so that the value is only expanded at the login step (at which point the value may have changed)

There was a small bug with #50 that I didn't catch until using the backend. When a downstream image (such as the component SDK or the AICA backend) creates an overlay workspace, redefining the `ROS2_WORKSPACE` variable, the colcon_cd path is not updated at login because the variable was already expanded directly in `bashrc`. Now it is written to file as the variable and only expanded at login.

### DIFF
--- a/ros2_ws/Dockerfile
+++ b/ros2_ws/Dockerfile
@@ -72,6 +72,9 @@ COPY --chown=${USER}:${USER} ./config/colcon ${COLCON_HOME}
 RUN /bin/bash ${COLCON_HOME}/setup.sh
 ENV COLCON_DEFAULTS_FILE ${COLCON_HOME}/defaults.yaml
 
+# set the default colcon workspace path
+ENV COLCON_WORKSPACE=${ROS2_WORKSPACE}
+
 # build ROS workspace
 RUN mkdir -p ${ROS2_WORKSPACE}/src
 WORKDIR ${ROS2_WORKSPACE}
--- a/ros2_ws/config/colcon/setup.sh
+++ b/ros2_ws/config/colcon/setup.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # enable the colcon_cd utility
-echo "source /usr/share/colcon_cd/function/colcon_cd.sh; export _colcon_cd_root=${ROS2_WORKSPACE}" >> ~/.bashrc
+echo 'source /usr/share/colcon_cd/function/colcon_cd.sh; export _colcon_cd_root="${ROS2_WORKSPACE}"' >> ~/.bashrc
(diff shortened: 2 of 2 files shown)


domire8 (approved):

eeberhard:
While this is open, I will raise the point I mentioned in https://github.com/aica-technology/backend/pull/30 again. I am using the `ROS2_WORKSPACE` variable to refer to the active colcon workspace, but we also have a `~/ros2_ws` base workspace. When we create overlay workspaces in the backend and component sdk, we actually put them into a `WORKSPACE` env variable for easier referencing throughout the dockerfile. 

This means we have inconsistent patterns in the base dockerfile compared to downstream ones:

```dockerfile
## in the ros2_ws Dockerfile ##
# create base workspace
ENV ROS2_WORKSPACE /home/${USER}/ros2_ws
RUN mkdir -p ${ROS2_WORKSPACE}/src
...

## in the backend Dockerfile ##
# create a new workspace
ENV WORKSPACE /home/${USER}/backend_ws
ENV ROS2_WORKSPACE ${WORKSPACE}  # set the new global workspace variable
RUN mkdir -p ${WORKSPACE}/src
...
```

But, maybe there is a better env variable we can that is less confusing. For example, would `COLCON_WORKSPACE` be better? Then it's maybe a little more verbose, but I think more explicit. 

```dockerfile
## in the ros2_ws Dockerfile ##
# create base workspace
ENV ROS2_WORKSPACE /home/${USER}/ros2_ws
RUN mkdir -p ${ROS2_WORKSPACE}/src
...
# set the default global workspace variable
ENV COLCON_WORKSPACE ${ROS2_WORKSPACE}
...

## in the backend Dockerfile ##
# create a new workspace
ENV WORKSPACE /home/${USER}/backend_ws
RUN mkdir -p ${WORKSPACE}/src
...
# overwrite the default global workspace variable
ENV COLCON_WORKSPACE ${WORKSPACE}
...
```

Let me know which one you prefer.

buschbapti:
I am good with `COLCON_WORKSPACE`. It might not be the most informative but I can't think of anything better and it will mostly be used internally / hidden to users so I can live with that.

domire8 (commented):
Works for me too

eeberhard:
Ok, thanks for the feedback. I will rename this variable to `COLCON_WORKSPACE`, and update it in the backend and component SDK (whenever a new overlay workspace is created) so that we can properly use `cb` and `ct` helpers in overlay workspaces.

PR #52, opened 2022-08-25
Fix COLCON_WORKSPACE path in colcon utils

eeberhard:
Well, I'm annoyed that this simple feature has caused about 6 PRs already, but there was one other bug I missed. As usual, it slipped under the radar when I was testing `cb` / `ct` on the `ros2_ws` image, but in the backend or other places with the overlay workspace it was always going to the base workspace.

To prevent unexpected behaviour, the default usage will still require you to `cd` into the appropriate workspace (overlay or underlay). When building overlay packages, the `-c | --cd` flag can be supplied to `cb` or `ct` so you can build from any directory.

@buschbapti @domire8

### DIFF
--- a/ros2_ws/config/colcon/build.sh
+++ b/ros2_ws/config/colcon/build.sh
@@ -9,10 +9,17 @@ whole workspace. Supply the following flags to modify the build.
 Any additional flags are forwarded to colcon build.
 
 -d | --debug            Set CMAKE_BUILD_TYPE to Debug.
+
 -o | --override         Add --allow-overriding for listed packages
+
+-c | --cd               Temporarily change into the COLCON_WORKSPACE
+                        directory when building
+                        (workspace path: ${COLCON_WORKSPACE})
+
 -h | --help             Show this help message.
 "
 
+CHANGE_DIR=false
 OVERRIDE=false
 PACKAGES=()
 BUILD_ARGS=()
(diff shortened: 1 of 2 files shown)


domire8:
Yeah I was going to let you know about that. But I still don't get what the expected usage would be. Can I now from any directory build any package of any workspace with `cb package_name` ?

eeberhard:
>Can I now from any directory build any package of any workspace with cb package_name ?

Yes, but only if you supply that extra `-c | --cd` flag. Unfortunately it doesn't work for packages in underlay workspaces like I originally hoped. It's a minor time-save but if you are in a container at `~/applications` and you want to rebuild a local component, you can just do `cb --cd <package>` instead of `cd ../workspace && cb <package> && cd ../applications`

domire8 (approved):
Okay...the downside of those changes is that you cannot do `cb` in the underlay workspace anymore which was quite nice because it was doing packages-select automatically.

eeberhard:
> Okay...the downside of those changes is that you cannot do `cb` in the underlay workspace anymore which was quite nice because it was doing packages-select automatically.

Well that's why I added the extra flag. If you manually `cd` to the underlay, you can still to `cb modulo_components` as usual. It's just that we can't automatically do it from anywhere.

To summarize:
1. If you are in a workspace, you can do `cb` or `cb <package_1> [<package_2> ... ]`
2. If you are in another directory and want to build a package in the *overlay* workspace, you can do `cb -c` or `cb -c <package_1> [<package_2> ... ]`
3. If you are in another directory and want to build a package in the *underlay* workspace, you have to manually `cd` to that workspace before doing step 1

domire8:
Ah perfect then!

PR #53, opened 2022-08-31
Add humble images to workflows

domire8:
I'm using here the same pattern for a two-way Dockerfile for ros2-modulo-control that we had before with foxy. After, I added all the humble-develop images to the relevant workflows so we get the latest develop images all the time. I find the main images unnecessary for now because they don't change often and we can build them ourselves if needed.

### DIFF
--- a/.github/workflows/build-push.yml
+++ b/.github/workflows/build-push.yml
@@ -236,7 +236,7 @@ jobs:
   build-publish-galactic-modulo-control-devel:
     needs: build-publish-galactic-modulo-devel
     runs-on: ubuntu-latest
-    name: Build and publish ROS2 galactic workspace image with modulo and ros2 control installed
+    name: Build and publish development ROS2 galactic workspace image with modulo and ros2 control installed
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
--- a/.github/workflows/build-push.yml
+++ b/.github/workflows/build-push.yml
@@ -41,6 +41,22 @@ jobs:
           ci_branch: ${{ env.CI_BRANCH }}
           secret: ${{ secrets.GITHUB_TOKEN }}
 
+  build-publish-humble-workspace:
+    needs: create-new-ci-branch
+    runs-on: ubuntu-latest
+    name: Build and publish ROS2 humble workspace image
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Build and Push
+        uses: ./.github/actions/build-push
+        with:
+          workspace: ros2-ws
+          base_tag: humble
+          ci_branch: ${{ env.CI_BRANCH }}
+          secret: ${{ secrets.GITHUB_TOKEN }}
+
   build-publish-galactic-control-libraries-devel:
     needs: build-publish-galactic-workspace
     runs-on: ubuntu-latest
@@ -69,6 +85,36 @@ jobs:
         with:
           hash: ${{ env.CL_DEVELOP_HASH }}
           file: ./control-libraries-develop-hash
+          ci_branch: ci
+
+  build-publish-humble-control-libraries-devel:
+    needs: build-publish-humble-workspace
+    runs-on: ubuntu-latest
+    name: Build and publish development ROS2 humble workspace image with control libraries installed
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Build and Push
+        uses: ./.github/actions/build-push
+        with:
+          workspace: ros2-control-libraries
+          base_tag: humble
+          cl_branch: develop
+          output_tag: humble-devel
+          ci_branch: ${{ env.CI_BRANCH }}
+          secret: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Get control libraries develop hash
+        run: |
+          HASH=$(git ls-remote https://github.com/epfl-lasa/control-libraries.git develop | awk '{ print $1 }')
+          echo "CL_DEVELOP_HASH=${HASH}" >> $GITHUB_ENV
+
+      - name: Write hash
+        uses: ./.github/actions/write-hash
+        with:
+          hash: ${{ env.CL_DEVELOP_HASH }}
+          file: ./control-libraries-humble-develop-hash
           ci_branch: ci
 
   build-publish-galactic-control-libraries:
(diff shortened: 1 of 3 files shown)


ros2_modulo_control/Dockerfile
```diff
@@ -17,6 +16,26 @@ RUN git clone -b galactic --depth 1 https://github.com/ros2/rcl_interfaces.git
...
+RUN git clone -b ros2-master --depth 1 https://github.com/ros-controls/control_toolbox.git
+RUN git clone -b 2.2.0 --depth 1 https://github.com/ros-controls/realtime_tools.git
+RUN git clone -b humble --depth 1 https://github.com/ros2/rcl_interfaces.git
+RUN git clone -b humble --depth 1 https://github.com/ros2/test_interface_files.git
+RUN git clone -b ros2 --depth 1 https://github.com/ros/angles.git
+RUN git clone -b ros2 --depth 1 https://github.com/ros-drivers/ackermann_msgs.git
+
+
+ARG BASE_TAG=galactic
+FROM ros2-control-version-${BASE_TAG} AS final
```
eeberhard, line 38: Minor style comment
```suggestion
FROM ros2-control-version-${BASE_TAG} as final
```

ros2_modulo_control/Dockerfile
```diff
@@ -1,13 +1,12 @@
 ARG BASE_IMAGE=ghcr.io/aica-technology/ros2-modulo
 ARG BASE_TAG=galactic
 
-FROM ${BASE_IMAGE}:${BASE_TAG}
+FROM ${BASE_IMAGE}:galactic as ros2-control-version-galactic
```
eeberhard, line 4: It seems this would discard all `galactic-devel` images in our final image layer... unfortunately I think you'll need to split the layers for each base tag. To avoid duplication, you can maybe do something along the following lines:

```dockerfile
FROM ghcr.io/aica-technology/ros2-ws:galactic as ros2-control-sources-galactic
USER ${USER}

RUN mkdir -p ${ROS2_WORKSPACE}/src/ros2_control
WORKDIR ${ROS2_WORKSPACE}/src/ros2_control

RUN git clone ...
...


FROM ${BASE_IMAGE}:galactic as ros2-control-version-galactic
COPY --from=ros2-control-sources-galactic ${ROS2_WORKSPACE}/src/ros2_control ${ROS2_WORKSPACE}/src/ros2_control

FROM ${BASE_IMAGE}:galactic-devel as ros2-control-version-galactic-devel
COPY --from=ros2-control-sources-galactic ${ROS2_WORKSPACE}/src/ros2_control ${ROS2_WORKSPACE}/src/ros2_control
```

ros2_modulo_control/Dockerfile
```diff
@@ -17,6 +16,26 @@ RUN git clone -b galactic --depth 1 https://github.com/ros2/rcl_interfaces.git
 RUN git clone -b galactic --depth 1 https://github.com/ros2/test_interface_files.git
 RUN git clone -b ros2 --depth 1 https://github.com/ros/angles.git
 
+
+FROM ${BASE_IMAGE}:humble as ros2-control-version-humble
```
eeberhard, line 20: Same issue here, though we could stick with only `humble-devel` for now. Ultimately I think the same approach should be taken as with galactic / galactic-devel

eeberhard (commented):
Nice work but I think there are some problems

ros2_modulo_control/Dockerfile
domire8: Good catch, thanks!

ros2_modulo_control/Dockerfile
```diff
@@ -17,6 +21,34 @@ RUN git clone -b galactic --depth 1 https://github.com/ros2/rcl_interfaces.git
 RUN git clone -b galactic --depth 1 https://github.com/ros2/test_interface_files.git
 RUN git clone -b ros2 --depth 1 https://github.com/ros/angles.git
 
+
+FROM ${BASE_IMAGE}:galactic as ros2-control-version-galactic
```
eeberhard, line 25: Remember that an ARG is persisted between stages. You will need to re-declare it each time for every stage.
```suggestion
ARG BASE_IMAGE=ghcr.io/aica-technology/ros2-modulo
FROM ${BASE_IMAGE}:galactic as ros2-control-version-galactic
```
It's worth the code duplication to keep the `--local-base` option imo

ros2_modulo_control/Dockerfile
```diff
@@ -1,13 +1,17 @@
 ARG BASE_IMAGE=ghcr.io/aica-technology/ros2-modulo
 ARG BASE_TAG=galactic
 
-FROM ${BASE_IMAGE}:${BASE_TAG}
+FROM ${BASE_IMAGE}:galactic as ros2-control-version-galactic
 USER ${USER}
-ARG ROS2_CONTROL_TAG=1.0.0
```
eeberhard, line 6: All of these lines can be removed now I think. The layer
```dockerfile
FROM ${BASE_IMAGE}:galactic as ros2-control-version-galactic
```
is declared correctly later so isn't needed here. The ARGs for BASE_IMAGE and BASE_TAG are also not consumed in that initial stage so can be removed.

ros2_modulo_control/Dockerfile
```diff
@@ -17,6 +21,34 @@ RUN git clone -b galactic --depth 1 https://github.com/ros2/rcl_interfaces.git
...
+
+FROM ${BASE_IMAGE}:galactic as ros2-control-version-galactic
+COPY --from=ros2-control-sources-galactic ${ROS2_WORKSPACE}/src/ros2_control/ ${ROS2_WORKSPACE}/src/
+
+
+FROM ${BASE_IMAGE}:galactic-devel as ros2-control-version-galactic-devel
+COPY --from=ros2-control-sources-galactic ${ROS2_WORKSPACE}/src/ros2_control/ ${ROS2_WORKSPACE}/src/
+
+
+FROM ${BASE_IMAGE}:humble as ros2-control-version-humble-devel
```
eeberhard, line 33: This should include `-devel` I think
```suggestion
ARG BASE_IMAGE
FROM ${BASE_IMAGE}:humble-devel as ros2-control-version-humble-devel
```

eeberhard (commented):
Seems like there are still some kinks. This is worth testing on a fork before merging.

eeberhard:
I was wrong about the ARG before each layer. Any ARG used in a FROM statement have to be declared at the very top. I've been testing some variants on a fork of this branch and hope to push some commits here to close it today. I didn't think you'd be free to work on it today

domire8:
https://github.com/domire8/docker-images/actions/runs/3021436641

eeberhard (commented):
Thanks for the revisions, I also tested them and it seems great!

eeberhard (approved):

.github/workflows/build-push.yml
```diff
@@ -129,6 +175,35 @@ jobs:
...
+          echo "MODULO_DEVELOP_HASH=${HASH}" >> $GITHUB_ENV
+
+      - name: Write hash
+        uses: ./.github/actions/write-hash
+        with:
+          hash: ${{ env.MODULO_DEVELOP_HASH }}
+          file: ./modulo-humble-develop-hash
+          ci_branch: ci
+
   build-publish-galactic-modulo:
```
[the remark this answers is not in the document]
domire8: @eeberhard thanks for the attention to detail

PR #55, opened 2022-09-13
Use humble branch for ros2-modulo:humble-devel image

domire8:
As title says, this should not need a fork, I'm fairly confident this is going to do what we expect.


eeberhard (approved):
Seems good

PR #56, opened 2022-09-14
Support ARM64

eeberhard:
This change will enable us to run our ROS stack on ARM platforms!

Downsides: The buildx pipeline takes _much_ longer because of QEMU; it emulates ARM architecture on an AMD server. This means many hours (10-12) to build all images, as opposed to the current 30 minutes or so.

I think this is acceptable if the job only runs once per day, and only when there are changes to the respective hashes of CL or modulo. However, it will not be conducive to rapid testing of changes on the develop branch. 

I see two options:
1. either we build platform-specific development images locally as needed (using the build scripts with --local-base),
2. or we have a separate workflow that just builds amd64 images as before, under a variant tag (`:galactic-amd64`, e.g. `:galactic-devel-amd64`)

I prefer option 1, because it means you can build the correct image for your own platform, and local builds will generally be faster than a CI build anyway. To make that process easier I have also included a `build-local.sh` script.

Depends on:
- https://github.com/epfl-lasa/control-libraries/pull/310

TODO before closing: revert control libraries branches from `feature/arm64` to `develop`, and re-run epfl-lasa control-libraries so that the multi-arch base deps support 20.04 images and upwards


domire8:
I have to say, I don't agree with a CI that takes 12h to complete and thus runs for half of the time of every day. God knows where the resources for that is coming from, we are not even paying for it, etc. 
If you think it's absolutely necessary in this way, go for it but I don't really support it. Maybe we can just run it once a week or something?

eeberhard:
> I think this is acceptable if the job only runs once per day, and only when there are changes to the respective hashes of CL or modulo

I was considering that CL and modulo are a bit more stable now, and the develop branches don't change every day. But still you're right that it's not necessary to always build every development image. I will restore the original workflow (to build the amd64 images daily when the repos are updated), and keep this workflow separate to be only once per week (if repos have changed) or on manual dispatch

domire8 (approved):
That, I can fully support :+1: Thanks, I think for the time being thats a reasonable compromise

eeberhard:
Summary of the new workflow behavior:

## Build and push workflow:
On manual dispatch or on push to `docker-images` main branch, build all multi-arch images for amd64 and arm64
- `galactic` [amd64 & arm64]
- `galactic-devel` [amd64 & arm64]
- `humble-devel` [amd64 & arm64]

## Scheduled workflows:

Main branches are rebuilt for amd64 and arm64 on bi-weekly workflows
- `galactic` [amd64 & arm64]

Develop branches are built _for amd64 only_ on "nightly" schedule (actually twice per day)
- `galactic-devel-nightly` [amd64]
- `humble-devel-nightly` [amd64]

domire8:
Thanks, that's what we discussed last week and it looks great. Could you support the two Dockerfiles in the build script too, please?

domire8 (approved):

eeberhard:
Tested on [my fork](https://github.com/eeberhard/docker-images/actions) over the weekend, and it's behaving as expected. The galactic main build currently fails because, on arm64, it cannot install Pinocchio from robot-pkg. The new install script on develop branch works on ARM, and the next CL release will resolve that issue.

PR #61, opened 2022-10-24
Remove config directory before committing

domire8:
Ok I'm getting a bit desperate here, the previous build and push workflow didn't succeed for the development modulo images and I can't really figure out why. While the build-push action works perfectly fine for most images, it fails sometimes, saying that there are no tracked changes but untracked files despite there being a commit on the CI branch with the desired hash.

So something else we can try is to remove the config directory again after the build stage such that it will not become an untracked change...


eeberhard (approved):
Seems reasonable

PR #62, opened 2022-10-28
Build galactic, humble and humble-devel images

domire8:
Okay so this is kinda the 'release' of humble. We replace the galacitc-devel image with a humble-devel image. This means:
- Galactic image with main branch of control libraries, galactic branch of modulo (which is equivalent to v2.1.1)
- Humble image with main branch of control libraries, main branch of modulo
- Humble-devel image with main branch of control libraries, humble branch of modulo (TODO!! this shall be changed to develop once we merged humble into develop on modulo)

I'm currently running this here: https://github.com/domire8/docker-images/actions/runs/3344142806

### DIFF
--- a/.github/workflows/check-upstream-devel.yml
+++ b/.github/workflows/check-upstream-devel.yml
@@ -17,8 +17,6 @@ jobs:
     outputs:
       cl_id: ${{ steps.check_cl.outputs.id }}
       cl_rebuild: ${{ steps.check_cl.outputs.rebuild }}
-      modulo_develop_id: ${{ steps.check_modulo_develop.outputs.id }}
-      modulo_develop_rebuild: ${{ steps.check_modulo_develop.outputs.rebuild }}
       modulo_humble_id: ${{ steps.check_modulo_humble.outputs.id }}
       modulo_humble_rebuild: ${{ steps.check_modulo_humble.outputs.rebuild }}
     steps:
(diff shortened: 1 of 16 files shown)


buschbapti (approved):
This pipeline is getting nicely

.github/workflows/check-upstream-devel.yml
```diff
@@ -17,8 +17,6 @@ jobs:
     outputs:
       cl_id: ${{ steps.check_cl.outputs.id }}
       cl_rebuild: ${{ steps.check_cl.outputs.rebuild }}
-      modulo_develop_id: ${{ steps.check_modulo_develop.outputs.id }}
-      modulo_develop_rebuild: ${{ steps.check_modulo_develop.outputs.rebuild }}
       modulo_humble_id: ${{ steps.check_modulo_humble.outputs.id }}
       modulo_humble_rebuild: ${{ steps.check_modulo_humble.outputs.rebuild }}
```
eeberhard, line 21: I think renaming these will also be part of the TODO you mentioned (when modulo develop branch officially supports humble). But for now it's more clear to keep them named like this.

Either it becomes `cl_develop_id` and `modulo_develop_id`, or back to just `cl_id` and `modulo_id`, whichever is cleaner. Since they are just output flags, the latter is probably better and more consistent with the other workflow.

eeberhard (approved):
Looks good!

.github/workflows/check-upstream-devel.yml
domire8: Indeed yeah, and good suggestion :+1:

domire8:
https://github.com/domire8/docker-images/actions/runs/3344668867

This seems to work, I will merge once control libraries patch release is merged as well, to save at least some computation

PR #63, opened 2022-10-30
Copy control libraries licenses to AICA images

domire8:
Now that we have this sorted out, I think we should also copy the licenses from the control libraries image.

Another option would be to include the "installing" of the licences with the install script but that would take some time again until its on main branch of control libraries
@eeberhard


buschbapti (approved):
Looks good. As discussed I also think it would make sense that the install script of control-libraries also install those licenses.

PR #64, opened 2022-11-09
Fix the lookup of the image hashes

domire8:
Right before the release of component SDK, I realized that the CI has another flaw. Since we started using `docker buildx build`, the hash of the image was never correctly retrieved and saved to file anymore. We need to use `docker buildx imagetools inspect` to get the hash of those multi arch files.

The changes are currently tested on my fork: https://github.com/domire8/docker-images/actions/runs/3428680106


domire8:
I stopped the workflow after all 3 ros2-control-libraries images were built, and the hashes were correctly written to file for all images

eeberhard (approved):
Nice find and fix. Seems good to me 👍

PR #65, opened 2022-11-10
Use develop branch for modulo humble-devel image

domire8:
Now that the humble branch is not available anymore, we need to use the develop one.

I also took the opportunity to merge the two build-push actions into one for reduced code duplication. A next step would be to squash the last image.

This is successfully running on my fork: https://github.com/domire8/docker-images/actions/runs/3437559109


eeberhard (approved):
The merging of the build action seems very sensible, and the renaming TODOs seem to be addressed. Very good!

PR #66, opened 2022-11-10
Export BUILD_FLAGS as env variable

domire8:
Sorry, one line that I forgot to copy...


eeberhard (approved):
I missed it too, no worries. Let's get it patched 👍

PR #67, opened 2022-11-14
Scheduled check of multi-arch humble and humble-devel

domire8:
As discussed, the nightly images were favoring people with AMD arch and resulted in the humble-devel images never being updated automatically. Let's shift back to upstream checks of humble (once per week) and humble-devel (twice per week) images and rebuild them for both AMD and ARM.

I also managed to reduce code duplicated with a reusable workflow. This was all tested successfully on my fork


eeberhard (approved):
I agree with the strategy as discussed. Seems good as far as I can tell.

PR #68, opened 2022-11-21
Reduce number of layers

domire8:
I was also working on the docker images over the weekend @eeberhard :smile: 
Due to the fact that the stash option is deprecated with buildx and will most likely not come back, we need to leverage multi-stage builds to reduce the number of layers in our Dockerfiles. With the proposed changes, I was able to reduce the amount of layers from 66 to 17 in the ros2-modulo-control image. How, you ask?

Use intermediate stages to build the sources and copy only the results (results in one additional layer only, especially easy for modulo and ros2 control, because it's just the ros2 workspace that was changed)

For the workspace and control libraries, it's slightly more complicated..
- For the ros2-ws, I need a layer that defines all the environment variables on the ros:humble image, then I copy the whole filesystem in the last stage.
- For ros2-control-libraries, the install script creates a symbolic link for eigen. Because docker cannot copy and override a real file with a symlink, I need to remove `/usr/include/eigen` from the base image before the copy step.

I'm currently testing this image with the robot and I cannot spot any problems so far

### DIFF
--- a/ros2_control_libraries/Dockerfile
+++ b/ros2_control_libraries/Dockerfile
@@ -19,7 +19,7 @@ COPY --from=ghcr.io/epfl-lasa/control-libraries/development-dependencies:20.04 \
   /usr/share/doc/control-libraries/licenses /usr/share/doc/control-libraries/licenses
 
 
-FROM ${UBUNTU_VERSION}
+FROM ${UBUNTU_VERSION} as control-libraries-build
 WORKDIR ${HOME}
 
 RUN sudo ldconfig
@@ -38,3 +38,6 @@ RUN sudo pip3 install ./control-libraries/python
 RUN sudo rm -rf ./control-libraries
 
 RUN sudo apt-get clean && sudo rm -rf /var/lib/apt/lists/*
+
+FROM ${BASE_IMAGE}:${BASE_TAG} as final
+COPY --from=control-libraries-build /usr /usr
--- a/ros2_control_libraries/Dockerfile
+++ b/ros2_control_libraries/Dockerfile
@@ -1,6 +1,6 @@
 ARG BASE_IMAGE=ghcr.io/aica-technology/ros2-ws
 ARG BASE_TAG=humble
-ARG UBUNTU_VERSION=focal-fossa
+ARG UBUNTU_VERSION=jammy-jellyfish
 FROM ${BASE_IMAGE}:${BASE_TAG} as jammy-jellyfish
 
 # install development dependencies and copy licenses
--- a/ros2_control_libraries/Dockerfile
+++ b/ros2_control_libraries/Dockerfile
@@ -40,4 +40,5 @@ RUN sudo rm -rf ./control-libraries
 RUN sudo apt-get clean && sudo rm -rf /var/lib/apt/lists/*
 
 FROM ${BASE_IMAGE}:${BASE_TAG} as final
-COPY --from=control-libraries-build /usr /usr
+RUN sudo rm -rf /usr/include/eigen3
+COPY --from=control-libraries-build /usr/ /usr/
--- a/ros2_control_libraries/Dockerfile
+++ b/ros2_control_libraries/Dockerfile
@@ -35,9 +35,9 @@ RUN sudo ./control-libraries/protocol/install.sh -y
 
 # install python bindings
 RUN sudo pip3 install ./control-libraries/python
-RUN sudo rm -rf ./control-libraries
 
-RUN sudo apt-get clean && sudo rm -rf /var/lib/apt/lists/*
+# clean image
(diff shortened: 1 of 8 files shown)


ros2_ws/Dockerfile
```diff
@@ -1,15 +1,23 @@
 ARG BASE_TAG=humble
-FROM ros:${BASE_TAG} as base-dependencies
+FROM ros:${BASE_TAG} as environment-variables
 ENV DEBIAN_FRONTEND=noninteractive
 ENV PYTHONWARNINGS=ignore:::setuptools.command.install,ignore:::setuptools.command.easy_install,ignore:::pkg_resources
+ENV PIP_NO_CACHE_DIR 1
```
buschbapti, line 5: Could you define those in a config file and load it in one line? Would that help or actually not even change anything?

ros2_control_libraries/Dockerfile
```diff
@@ -38,3 +38,7 @@ RUN sudo pip3 install ./control-libraries/python
 RUN sudo rm -rf ./control-libraries
 
 RUN sudo apt-get clean && sudo rm -rf /var/lib/apt/lists/*
+
+FROM ${BASE_IMAGE}:${BASE_TAG} as final
+RUN sudo rm -rf /usr/include/eigen3
+COPY --from=control-libraries-build /usr/ /usr/
```
buschbapti, line 44: I am not sure I understand what is going here. Are you not still in the `control-libraries-build` layer?

ros2_modulo/Dockerfile
```diff
@@ -13,3 +13,6 @@ WORKDIR ${ROS2_WORKSPACE}
 RUN /bin/bash -c "source /opt/ros/${ROS_DISTRO}/setup.bash; colcon build"
 
 RUN sudo rm -rf /tmp/*
+
+FROM ${BASE_IMAGE}:${BASE_TAG} as final
+COPY --from=modulo-build ${ROS2_WORKSPACE} ${ROS2_WORKSPACE}
```
buschbapti, line 18: Same comment as above. I guess I don't really understand how that works.

buschbapti (commented):
That is quite interesting. Not sure I fully grasp (see comments). Seems ok otherwise.

ros2_control_libraries/Dockerfile
buschbapti: ok nevermind I missed the `FROM` line 42. All good now.

buschbapti: Could there be some conflicts between lib installed in the base ubuntu image and the ros image used line 42? How does the copy handles duplicates? Is that a recursive merge?

ros2_ws/Dockerfile
eeberhard: I don't think there's a clean way to do it at build time. On build, you can pretty much only set build-args externally.  We can COPY and `RUN source / export` from a file, but that variable will not be available to subsequent layers (unlike the ENV directive).

For docker run, there is an `--env-file` parameter. But that doesn't help us in the CI.

For the second part of the question, I think because ENV directives do not make any changes to the filesystem, they are treated differently as intermediate layers. So, I don't think it's too problematic to have them listed here.

ros2_ws/Dockerfile
```diff
@@ -89,15 +88,16 @@ RUN mkdir /root/.ssh/ && ssh-keyscan github.com | tee -a /root/.ssh/known_hosts
 
 # prepend the environment sourcing to bashrc (appending will fail for non-interactive sessions)
 RUN echo "source /opt/ros/${ROS_DISTRO}/setup.bash; \
-source ${ROS2_WORKSPACE}/install/setup.bash" | cat - ${HOME}/.bashrc > tmp && mv tmp ${HOME}/.bashrc
+  source ${ROS2_WORKSPACE}/install/setup.bash" | cat - ${HOME}/.bashrc > tmp && mv tmp ${HOME}/.bashrc
```
eeberhard, line 91: This will add the leading whitespace in the bashrc file - it's not really a problem but we can either format it nicely here in the Dockerfile or in the final file, and I would prefer the latter (how it was before).

eeberhard (approved):
Nice work, a lot of those simplifications were straightforward and overdue (when just copying the ros2 workspace), and the other changes for environment variables and Eigen issues are clever and neat.

If this is working smoothly then I am happy with it!

ros2_control_libraries/Dockerfile
domire8: I guess there could but we are only copying `/usr/local` from the ubuntu image of control libraries so I think it's unlikely that something happens there. Not sure what type of merge it is.

domire8:
I also did a last update on the ros images for the people at the lab that might use them. I will build and push them manually

PR #69, opened 2022-11-28
Remove galactic from workflows

domire8:
This PR had very low priority because I didn't mind the galactic builds very much but now, I realized with @Maxime00 that the `sudo ldconfig` is very much necessary at the end of the control libraries images, so this became now the actual reason for this PR.


eeberhard (approved):
Sure, looks good. Only thing is that the PR title is a bit misleading since there are actually three things going on:
- remove galactic from build push workflows
- configure control libraries post installation
- allow `noetic` base tag for images

If you can find a better commit title or description to make it clear when you close it, that would be great

PR #70, opened 2022-12-21
Use control libraries and modulo from AICA

domire8:
I believe a simple search and replace of epfl-lasa will do the trick here


eeberhard (approved):
Looks good to me!

Issue #72, opened 2023-02-03
GitHub registry lists unknown/unknown as architecture


domire8:
Basically this: https://github.com/orgs/community/discussions/45969

PR #73, opened 2023-02-03
Avoid unknown architecture under packages with additional build flags

domire8:
## Description

- #72 

This PR solves the issue by doing what is proposed in the resource. This is tested on my fork

## Review guidelines


eeberhard (approved):
Interesting, I hadn't seen that issue. Seems sensible based on the linked resources.

PR #74, opened 2023-02-08
Add reusable workflow for build-push

domire8:
## Description

- https://github.com/aica-technology/.github/issues/3

This PR adds a workflow that wraps the two workflows (multi arch build-push and write file) from our .github repository for the particular case of our docker-images. In other words, this translates - with minor differences - the [build-push action](https://github.com/aica-technology/docker-images/blob/main/.github/actions/build-push-multi-arch/action.yml) into a workflow, such that we are able to reuse other workflows from other places.

While we still have that special parsing of inputs that can get the Dockerfile path and workspace directory from the image names (`ros2_control_libraries` with branch `develop` if you want to build `ros2-control-libraries:humble-devel` etc.) we have the special case of the `ros2_ws` where we need to copy the `sshd_entrypoint.sh` script. For that, we need to upload the file as an artifact and pass the artifact name and path to the build-push workflow from the .github repo.

I want to show you the new workflow before I delete the existing actions and refactor the current workflows. This has been tested on my fork already.

## Review guidelines

Estimated Time of Review: x minutes

### DIFF
--- a/.github/workflows/reusable-build-push.yml
+++ b/.github/workflows/reusable-build-push.yml
@@ -24,10 +24,10 @@ on:
       dockerfile_extension:
         required: false
         type: string
-      architecture:
+      platforms:
         required: false
         type: string
-        default: 'amd64 & arm64'
+        default: 'linux/amd64,linux/arm64'
       write_hashes:
         required: false
         type: string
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-build-push.yml
```diff
@@ -0,0 +1,198 @@
+name: Reusable workflow for build and push multi-arch images
```
eeberhard, line 1:
```suggestion
name: Reusable workflow to build and push multi-arch images
```

.github/workflows/reusable-build-push.yml
```diff
@@ -0,0 +1,198 @@
...
+
+# Run workflow by manual dispatch
+on:
+  workflow_call:
```
eeberhard, line 5: You say manual dispatch in the comment but it's using the `workflow_call` event, which is triggered only by other workflows. Is the comment wrong or is the event wrong?

eeberhard (commented):
There's a lot going on but it seems pretty good. I think it's a good idea to put this as a separate workflow first, integrating it with refactored workflows and then removing the old steps later.

.github/workflows/reusable-build-push.yml
domire8: Yes the comment was wrong

eeberhard (approved):

domire8:
Needed to rename the `architecture` argument to `platforms

Issue #76, opened 2023-03-07
Black screen in RViz


domire8:
See
- osrf/docker_images/issues/662

The fix is
```
add-apt-repository ppa:kisak/kisak-mesa
apt update
apt upgrade
```

domire8:
@eeberhard agree that we had this issue on linux and not on Mac virtual machines? I can't fully remember

domire8:
This should now not be necessary anymore, so we can remove the lines again

PR #77, opened 2023-03-07
Fix black RViz screen

domire8:
## Description

- #76 

This PR solves the issue by applying the fix from the ros issue.

I wanted to add that line just below the other apt installs, but somehow it completely messes up permissions and then the home folder is owned by root instead of ros2. Please don't ask me why a `apt-add` would cause that... Tested on the virtual machine :+1:

## Review guidelines

Estimated Time of Review: 2 minutes, except if you want to find out why we can make that upgrade up top


ros2_ws/config/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
eeberhard, line 1: I assume you didn't mean to add this file to this PR?

eeberhard (changes requested):
Apart from the file from another feature in development, the fix does work for me as well 👍 seems fine there temporarily

ros2_ws/config/entrypoint.sh
domire8: Yes of course, thanks

eeberhard (approved):

Issue #78, opened 2023-03-13
Permission issues with UIDs different from 1000


domire8:
It is possible that some linux users have UID/GID different from 1000, e.g. when they are not the first user created on the system. This leads to permission issues with the filesystem because the `ros2` user of the aica-docker images has UID/GID 1000.

It would be nice to be able to set the UID and GID according to the local user that is creating the interactive session. The 'server' case is not affected by this, because we launch it with user `root`

domire8:
@LouisBrunner i'd like to make you aware of this issue here. Our current backend and application images work a lot with the filesystem of the host through volume sharing and if the UID/GID of the host and the docker user are not the same, things just get messed up. One way to solve this would be to not interact with the filesystem anymore but that seems like a long term goal. In the short term, we should somehow try to account for this.

LouisBrunner:
@domire8 Is it just the case to get users to set `--uid` and `--gid` when they start the image?

Otherwise, I imagine you can probably get away with forwarding logs through docker and maybe users don't need to have access to your DB/internal files anyway. I guess you could also get all configuration/file exports done in the Front-End and avoid this issue...

domire8:
> Is it just the case to get users to set `--uid` and `--gid` when they start the image?

No, because the docker images on docker-images are built with a user that has uid and gid 1000. There are also some files (example applications) that are copied into the backend image just at the end. If you then start the backend from a host user with 1001, all the files will have wrong ownership and editing/creating files on shared volumes will be problematic.

For now, what I do is the following ([at the very top of the backend image](https://github.com/aica-technology/backend/blob/1d3cfc3b57e7d7006fb88d4e8f0da8aca8d21819/Dockerfile#L6))
```
USER root
RUN usermod -u 1001 ${USER} && groupmod -g 1001 ${USER} && chown -R ${USER}:${USER} ${HOME}
```

LouisBrunner:
Good point, however this is usually only an issue when writing files. Do you need to write in a lot of different places when using ROS? Otherwise, you could standardize a single folder with all your data, e.g. `/data` and make sure that that one is always writable.

domire8:
Hm yeah that would be one possible solution but especially in development/testing we sometimes share volumes to specific packages in the ROS workspace such that we can change the source code on the host and directly rebuild inside the container. But maybe this issue is also more about improving our workflows for that particular situation than anything else..Because apart from that it would be easy to just say all the shared data is in one place

domire8:
[...]

PR #79, opened 2023-03-13
Set UID and GID of local user in interactive container

domire8:
## Description

- #78 

This PR solves the issue by installing the tool [fixuid](https://github.com/boxboat/fixuid) (MIT license) that is invoked in the entrypoint script to change the UID and GID of the docker `ros2` user to the same values as the local user. Unfortunately, this may take up to a minute for the ros2-modulo-control image due to the recursive change of ownership of the home folder (which is why I will add additional steps in the backend Dockerfile, such that we can do this on build and not on run)

## Supporting information
Tested on
- Mac
- Virtual ARM64 ubuntu on mac
- AMD64 ubuntu

## Review guidelines

### DIFF
--- a/common/entrypoint.sh
+++ b/common/entrypoint.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+set -e
+
+if [ $(id -u) -ne 0 ]; then
+  echo "Fixing UID, this can take a few moments..."
+  eval $( fixuid -q )
+fi
+
+# setup ros2 environment
+source "/opt/ros/$ROS_DISTRO/setup.bash" --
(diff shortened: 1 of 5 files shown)


domire8:
> how would you resolve this at build time for the backend? 

```
USER root
RUN usermod -u ${UID} ${USER} && groupmod -g ${GID} ${USER} && chwon -R ${USER}:${USER} ${HOME}
```

> And would that approach work for pre-built backends that we deploy as images to other users?

Not really, as the UID/GID of the end user is unknown. Either the user has to build a personalized backend with just the line above or we have to set it on runtime.

PR #80, opened 2023-03-13
Fix aica-docker installation for Mac

domire8:
## Description

This PR addresses the problem with the aica-docker installation on Mac. So far, we added `autoload bashcompinit; bashcompinit` which might not be a known command if you use zsh and should rather be enabled by the user itself on another line of the bashrc file. In other words, if the user didn't enable autocomplete in general yet, aica-docker is not going to change that.

## Supporting information
Would be nice if you could test that @eeberhard 

## Review guidelines

### DIFF
--- a/scripts/src/aica-docker-completion.sh
+++ b/scripts/src/aica-docker-completion.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 SCRIPTS=("connect" "interactive" "server")
-CONTAINERS=$(docker ps --format "{{ .Names }}")
-IMAGES=$(docker images --format "{{.Repository}}:{{.Tag}}")
+CONTAINERS=$(docker ps --format "{{ .Names }}" 2> /dev/null)
+IMAGES=$(docker images --format "{{.Repository}}:{{.Tag}}" 2> /dev/null)
 IMAGES=${IMAGES//"<none>:<none>"/}
 
 _aica_docker () {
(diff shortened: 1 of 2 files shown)


eeberhard (commented):
Works for me!

Another reason I didn't initially install auto completion is that the docker engine is not always started by default on Mac, and so loading the auto completion gives multiple "docker engine not running" errors on each new terminal opening.

But I would like to use autocompletion, so I would suggest to suppress the warning by piping errors to dev null in `aica-docker-completion.sh`. That additional fix is also tested and working and makes it nice to use.

```diff
--CONTAINERS=$(docker ps --format "{{ .Names }}")
--IMAGES=$(docker images --format "{{.Repository}}:{{.Tag}}")
++CONTAINERS=$(docker ps --format "{{ .Names }}" 2> /dev/null)
++IMAGES=$(docker images --format "{{.Repository}}:{{.Tag}}" 2> /dev/null)
```

domire8:
Nice, thanks for testing

eeberhard (approved):

scripts/src/aica-docker-completion.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 SCRIPTS=("connect" "interactive" "server")
-CONTAINERS=$(docker ps --format "{{ .Names }}")
-IMAGES=$(docker images --format "{{.Repository}}:{{.Tag}}")
+CONTAINERS=$(docker ps --format "{{ .Names }}" 2> /dev/null)
+IMAGES=$(docker images --format "{{.Repository}}:{{.Tag}}" 2> /dev/null)
```
eeberhard, line 5: Sorry I forgot to mention these same commands actually happen lower down too

eeberhard (approved):

Issue #81, opened 2023-03-27
Symlink issues with multiple users using aica-docker


domire8:
As the symlink at `usr/local/bin/aica-docker` points to files in the home directory of the user that installed it, it will not work for other users. Better solution than the symlink is to merge all required scripts into one file and make an actual copy to `/usr/local/bin`.

PR #83, opened 2023-03-27
Combine aica-docker scripts to one script to avoid symlinks

domire8:
## Description

- #81 

This PR solves the issue by prepending the 'connect', 'interactive', and 'server' scripts as functions to the 'aica-docker' script and then actually copy the latter to `/usr/local/bin`.

## Review guidelines

## Checklist before merging:
- [x] Test on NUC (linux)

### DIFF
--- a/scripts/aica-docker.sh
+++ b/scripts/aica-docker.sh
@@ -1,15 +1,8 @@
 #!/bin/bash
 
-function connect () {
-  source "${SCRIPT_DIR}"/src/connect.sh "$@"
-}
-
-function interactive () {
-  source "${SCRIPT_DIR}"/src/interactive.sh "$@"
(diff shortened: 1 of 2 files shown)


domire8:
This does not work on linux..

eeberhard:
Will be glad to test if it works on linux and Mac, and if so it's a nice fix for the issue

domire8:
I've tested this on mac, and on the NUC with two user accounts on linux, should be fine as far as I can tell

eeberhard (approved):
Tested on my side and works as well, thanks!

domire8:
Could you test again if that fixed your issue?

eeberhard (approved):

PR #87, opened 2023-06-07
Add realtime group and add ros2 user to it

domire8:
## Description

Add the ros2 user to a user group that has realtime permissions

## Review guidelines


eeberhard (approved):
It's difficult to think of a meaningful group ID, but it's important for both the docker container and the host system to have the same ID for the realtime group in order for RT stuff to work. So, this random and specific number will do 👍

PR #88, opened 2023-06-27
Enable builds with iron base images

domire8:
## Description

`ros:iron` is now available as docker image, so we should start supporting it. This just enables builds with the `iron` tags, nothing more.

## Review guidelines


domire8:
I built the images both locally and with the CI and it worked fine

eeberhard (approved):

Issue #86, opened 2023-05-30
Revise the sequence of images


eeberhard:
Currently, this repo builds four consecutive images:

- ros2-ws -> set up ros2 user and base tooling on top of official ros2 image
- ros2-control-libraries -> install control libraries
- ros2-modulo -> add modulo packages
- ros2-modulo-control -> add `ros2_control` packages

This can be revised in several ways. For one, `ros2_control` packages take a fair amount of time to build; they are maintained externally so we generally used a fixed version. For this reason, these packages could be built and installed in the base layer for more persistence.

A second issue is that control libraries (which are independent of ros2) are installed, from source, in a standalone image, from which the modulo image is built (also from source). The control libraries + modulo combination could be a single image (with a specific CL version + modulo version). By pre-building the modulo packages in the modulo CI, copying them into the final `ros2_modulo_control` image would be easier. 

The only question I have to the second point is how to match the "canonical" CL version in the final ros2 image with the CL version used to build the modulo package. For me it seems easier to "bundle" CL with modulo in the modulo CI. If you want CL version X and modulo version Y, that would be built on demand in modulo CI and then simply composed by the CI in this repo.

The diagram could look something like this:
![open-source-repos](https://github.com/aica-technology/docker-images/assets/32450951/a094dc17-10a9-4bb2-88cc-2c847784f896)

Not a huge priority, but in particular point 1 (moving ros2_control before CL and modulo) can save a fair bit of CI time in the rolling development builds.

Let me know your thoughts @LouisBrunner @domire8.

LouisBrunner:
I personally think that in the medium-term, we should do away completely with base images and keep a single one (`ros2-control` I imagine). Modulo/CL can be added dynamically when needed with whatever version is needed.

However, that requires more advanced dependency tree management so in the meantime this makes sense to me.

PR #89, opened 2023-06-27
Revise the sequence of images

domire8:
## Description

- #86 

This PR solves the issue by building a `ros2-control` image after the `ros2-ws` and then the control libraries and modulo on top. I know it doesn't solve the problem fully but as mentioned in the issue, its a good step anyway.

I tested this on my fork.

## Review guidelines


eeberhard (approved):
It's a good step and shouldn't break anything, let's go for it 👍

Issue #90, opened 2023-06-28
Migrate to Iron


eeberhard:
```[tasklist]
### Tasks
- [ ] #86
- [ ] #88
- [ ] https://github.com/aica-technology/dynamic-components/issues/267
- [ ] aica-technology/modulo#32
- [ ] https://github.com/aica-technology/dynamic-state-engine/issues/117
```

eeberhard:
@domire8 I think this one can also be closed 🔥

PR #93, opened 2023-07-17
Add ros2_control Dockerfile for iron

domire8:
## Description

- #90

This PR adds a Dockerfile for iron with updated tags for ros2 control and ros2 controllers (and all the additional annoying dependencies...).

## Review guidelines

Estimated Time of Review: x minutes


eeberhard (approved):
Good job hunting down those dependencies. Looks good for now and we can then see how it affects our DSE, hw interfaces and controllers

Issue #94, opened 2023-08-04
Future of base image


domire8:
I wanted to discuss with you @LouisBrunner and @eeberhard how you imagine the future of our base images here. Currently, there is the problem that the base images are not versioned (related to #82) so if the CI rebuilds the `iron-devel` image for example, the old one is lost.

Now that we also build control libraries and modulo, is the idea that we just package them together similar to the backend-base and have a tagged ros2_control image that will most likely not change until the next version of ros?

LouisBrunner:
In my mind, and as you said, I would see the images in this repository becoming pure ros2/ros2_control images. Control-libraries and Modulo would then be added later on a need-to basis (and in the final `base` image).

On the point of versioning, I would still version the images we produce but it doesn't mean we can't have a more generic tag as well.
For example: let's say we define the first Iron image as `1.0.0-iron` and `iron` and then increase the minor/patch version when we make changes, e.g. `1.1.0-iron` and `iron`. In the end we always have a latest `iron` but we can also pin exact builds. Once we upgrade to the next one we can just increase the major tag, e.g. `2.0.0-jjj`.
If the packages available or configuration setup changes in any way in between builds I think that would be better instead of a single rolling tag. In this case, we get both convenience and exactness.

On the point of `devel`, do we still need it if we don't have control-libraries bundled?

domire8:
Okay, that would make sense I think.

We don't need the devel anymore no.

eeberhard:
Yes, agreed with all points above. A generic tag for the base ros2_control image would be nice, followed by including the CL and Modulo dependencies as separately versioned images (thus removing those images from these workflows). Both of those are not blocking / high-priority but would certainly be nice to pick up in the next months when convenient.

domire8:
I need a bit more clarification on one point:
Do we still pre-build a `ros2-modulo-control` image here (triggered manually with specified CL and modulo tags) for all our packages that depend on that or is there no image like that and each package needs to copy the dependencies in its own image? In other words, will for example the DSE use `ros2-modulo-control` as base or `ros2-control` and then copy the colcon workspace from `modulo`?

I would prefer option 1, but maybe that was already what you suggested.

eeberhard:
The idea would be to build the output of `modulo` first and then copy the build output into the layers using `ros2-control` as base. This would make it much more flexible to change modulo version.

I imagine most dockerfiles would look something like:

```dockerfile
[...]

Issue #82, opened 2023-03-27
Add version to docker-images


domire8:
As title suggests. Best option would be a workflow that updates the version on push to main.

domire8:
comment from @eeberhard 
> As a side note, how do you feel about tagging the ros2-control image with the version (for example, as :iron and :iron-3.18)? That way if we keep iron but upgrade ros2_control to some other version later, we can distinguish them. Not in scope for this PR but goes in a similar direction and I think would be useful to do.

comment from @LouisBrunner 
> I think we should just make our own version for this. e.g. v2.3.0+iron

comment from me:
> Fair suggestion, do we take the version from a VERSION file?

LouisBrunner:
I don't mind where we get the version at the moment, as long as we produce 2 tags (`v2.3.0+iron` and `v2.3.0`, as it makes package management easier later). I don't mind which of these tags is the ones that makes it into git either.

LouisBrunner:
Iron and humble would be different versions, e.g. `v1.4.5+humble` and `v2.3.1+iron`. There is no reason for them to be linked (after all they aren't necessarily compatible).

domire8:
A couple of weeks ago, we decided that in the near future, we'll drop the `ros2-control` image and only ship one `ros2-ws` image that contains all dependencies we need for ROS2 and ROS2 Control. @LouisBrunner What would be your suggestion for the tags of the image? `vX.Y.Z+ROS_DISTRO`? I think we could already start to set that up.

LouisBrunner:
So the new `ros2-ws` will be equivalent to the old `ros2-control`?

Yeah I think a version like that makes sense.

domire8:
Name is not 100% sure but yeah essentially that's it.

domire8:
The `+` is not allowed in docker images tags though it seems

LouisBrunner:
Ah, that's a shame. I guess we can do multi-tags and include `vX.Y.Z`, `vX.Y.Z-ROS_DISTRO` and `ROS_DISTRO` and that should be both standard Docker and mostly semver-compliant.

PR #97, opened 2023-09-11
ci: only build ros2-ws and ros2-control

domire8:
## Description

- #94 

This one hurts a bit because it removes a lot of code that took some effort to make it work hehe. But as discussed in the topic, we only want to build ros2-ws and ros2-control images and then just add control-libraries and modulo as desired. For that we can remove all the workflows that check if those two repos have new changes. Additionally, there is now also a build with a manifest merge now, such that all pushed images have a proper tag.

## Review guidelines

## Checklist before merging:


LouisBrunner:
I think we should just make our own version for this. e.g. `v2.3.0+iron`

domire8:
Let's continue discussing in #82 . Is this one good to go?

Issue #98, opened 2024-04-23
Consolidate images into one base image


domire8:
As discussed as part of a different initiative (public base class for modulo controllers), it was agreed that there should only be one `ROS2` base image that contains all the sources for ROS2 and ROS2 Control. This has several advantages, mainly that we don't need to keep track of an additional intermediate image and that then all our repositories using ROS can use a tagged version of that one single image (see also #82).

I don't think this needs to be an epic, but the steps would be:
1. Tag the current state of the repo as v1.0.0 (done)
2. Remove Dockerfiles of deprecated images (not the images on GHCR itself!)
3. Merge Dockerfiles of `ros2-ws` and `ros2-control`
4. Update README
5. Release the new base image with the current tags of ros2-control as `v1.0.0+iron`

@eeberhard @bpapaspyros @LouisBrunner

LouisBrunner:
Just a reminder: as discussed in #82 we should have multiple tags for each release for maximum discoverability.

PR #100, opened 2024-04-25
refactor: merge ros2-control image into ros2-ws

domire8:
## Description

- #98 

Next part of the issue: Merging the ros2 control dockerfile with the one and only `ros2-ws` image. I used a little trick with the stages and `BASE_TAG` that would potentially allow a build of different ROS distros if we added for example a `ros2-control-jazzy` stage.

Next up: reworking the CI.

## Review guidelines

## Checklist before merging:


eeberhard (approved):
pretty straightforward, nice!

PR #101, opened 2024-04-26
refactor: build push workflow for ros2-ws

domire8:
## Description

- #98 

Here my first suggestions for the CI: One build-push workflow that rebuilds the images either on version change (see `VERSION.<ros-distro>` file) or if manually dispatched and forced. The resulting image tags are `<ros_distro>-vX.Y.Z` and `<ros_distro>`. I will test it out a bit on my fork.

## Review guidelines

## Checklist before merging:


domire8:
My tests show success

LouisBrunner (approved):

.github/workflows/reusable-build-push.yml
```diff
@@ -3,71 +3,89 @@ name: Reusable workflow to build and push multi-arch images
...
-          WS_PATH=${{ inputs.workspace }}
-          WS_PATH=${WS_PATH//[-]/_}
-          echo "path=${WS_PATH}" >> $GITHUB_OUTPUT
-          FILE=${WS_PATH}/Dockerfile${{ inputs.dockerfile_extension }}
-          echo "file=${FILE}" >> $GITHUB_OUTPUT
-
           echo "build_flags=${BUILD_FLAGS[@]}" >> $GITHUB_OUTPUT
 
+          VERSION_TAG="v${{ needs.check-version.outputs.version }}"
+          IMAGE_TAGS="${{ inputs.ros_distro}}-${VERSION_TAG}","${{ inputs.ros_distro}}"
```
LouisBrunner, line 77: Shouldn't this be:
```suggestion
          IMAGE_TAGS="${VERSION_TAG},${VERSION_TAG}-${{ inputs.ros_distro}}","${{ inputs.ros_distro}}"
```
So we get `1.0.0`, `1.0.0-iron` and `iron`?

domire8: I'm a bit afraid to have tags without the ROS distro in it. What if we have other distros at some point?

LouisBrunner: We should never reuse the tags even between distros anyway. What's worrying you?

domire8: Well if we always include the distro with the version, then we could potentially have a `1.0.0-jazzy` no? Given your statement it seems not though, so we can commit your suggestion if you prefer.

LouisBrunner: Jazzy would be a breaking change so we would go to `2.0.0-jazzy` right? For me the distro name is just a convenience for clarity (thus why we should have been using `+iron` but it's not possible so that's the next best thing).

PR #105, opened 2024-05-01
fix: correct user

domire8:
## Description

The newly generated `iron` image had problems with permissions since the root user would build the ros2 control packages. This should fix it

## Review guidelines

## Checklist before merging:


bpapaspyros (approved):
thanks!

bpapaspyros (approved):
ok for me if it's tested

PR #108, opened 2024-05-02
Fix ros2-ws image

domire8:
## Description

As described by @eeberhard:
> so just to clarify, the point is to re-tag the existing (and at this point unchanging) ros2-control image as ros2-ws with the addition of the metadata labels. Then afterwards, we can revert ros2-ws Dockerfile to fully build the base workspace and install ros2-control deps all in one. And we are doing that because rebuilding ros2-ws from scratch now pulls updated (and breaking) changes to ros2 iron itself.

## Review guidelines

## Checklist before merging:


domire8:
So we just have one problem now, the CI doesn't allow downgrading the version...We can merge this and then I can force the CI from a different branch.

bpapaspyros (approved):

ros2_ws/Dockerfile
```diff
@@ -1,152 +1,10 @@
...
-COPY --from=ros2-control / /
-
-# start as ROS user
-USER ${USER}
-WORKDIR ${ROS2_WORKSPACE}
-
-# Metadata
-ARG VERSION
-ARG BASE_TAG
+ARG VERSION=v0.0.0
```
eeberhard, line 4: Just a question without going into the CI workflow to check myself: the VERSION.iron is written as `1.0.0` (not `v1.0.0`) but here the default ARG includes the v prefix `v0.0.0` (which is something I just wrote for testing). Should the `org.opencontainers.image.version="${VERSION}"` include the v or not? I would just prefer to double check that the default format of the ARG is representative of the intended metadata format, since that helps readability and avoids confusion.

domire8: The version that the CI action from our `.github` repo gives is in the format `vX.Y.Z`. We never put the `v` in our `VERSION` files, since the package.xml also doesn't contain it. However, if Louis wants the `v` in the metadata, I can't be 100% sure but I suppose so since he made those changes himself.

eeberhard (approved):

ros2_ws/Dockerfile
LouisBrunner: For future reference:

* For git labels: I find it convenient to have them as it does better grouping of the tags
* For anything we parse (aica-pacakage.toml, metadata): with or without doesn't matter as our parser handle either (https://github.com/hashicorp/go-version/blob/v1.6.0/version.go#L21)

PR #109, opened 2024-05-10
feat: (reverts #108) full ros2-ws Dockerfile with fixed SHA256 for ros:iron

eeberhard:
## Description

The last PR #108 reverted three previous commits and replaced the `ros2-ws` Dockerfile with a simple wrapper on a previously tagged version of the (now legacy) `ros2-control` image. It was merged as a hotfix because the underlying `docker.io/library/ros:iron` image has been rebuilt and retagged many times since our last build of `ros2-control`, so the most recent build of `ros2-ws` with the plain `ros:iron` tag actually included breaking changes to the `rclcpp::Node` API and therefore breaking downstream users like the `modulo` library. Using the old `ros2-control` image that was last built in September 11, 2023 was a temporary workaround so that we could continue refactoring downstream users to expect the new holistic `ros2-ws` image instead of the previous variants like `ros2-control-libraries`, `ros2-modulo` or `ros2-control`.

Now that the initial hotfix has had time to settle in, I would prefer to revert the revert, i.e. to go back to the intended future structure of this Dockerfile and repository. However, we still want to avoid the breaking change in the base `ros:iron` image (for now). This can be achieved using a specific SHA256 digest instead of the plain tag.

The repository info for official docker images (https://github.com/docker-library/repo-info) is updated with the SHA256 digest of all builds. By going back in history to the ROS [tag-details file from September 02, 2023](https://github.com/docker-library/repo-info/blob/6ed8f54531fe47e91d0da7411d77eb573ca08e69/repos/ros/tag-details.md#rosiron), we can identify the SHA256 digest of our desired base ros:iron image as:

```
ros@sha256:6df9b084cb7e918455df628126b2b647d2ad2e4b0e979fb8fbc525a610573bbb
```

I updated the Dockerfile and build workflow in commit 7d98f0a716dba1a58fe5fdf9819af8c7e3e36a66 to use more build args (ROS_DISTRO, BASE_IMAGE, BASE_TAG) and some conditional logic to set the specific digest for `iron` while using the normal tag for other distros.

As part of this followup work, in commit 4bf3b7b4708c989b05f32e3b331ec5cbe939a07c I also fixed an issue with the image metadata. The tech.aica.image.metadata base.version currently uses BASE_TAG (which is equivalent to the ROS distro, i.e. iron)). I realized that `package-builder:v1` will fail to parse the metadata since it expects the base version as a semver string. We will get an error along the lines of:

```console
ERROR: failed to solve: invalid AICA metadata "{\"type\":\"base/ws\",\"base\":{\"name\":\"docker.io/library/ros\",\"version\":\"iron\"}}" in image "@aica/foss/ros2-ws": Malformed version: iron
```

In my opinion the simplest and safest approach is to use the VERSION arg instead (which is set from the git tag, i.e. `vX.Y.Z-foo`). This way the version is parsed as valid, and we can still introspect the ROS distro from it. As far as I am aware, the exact version or formulation of `LabelAICABase.Version` is not critical as long as we are consistent with it, but please correct me if I am mistaken @LouisBrunner.

Finally, while adjusting the image metadata, I also added a label for devcontainer metadata (see more info at https://containers.dev/implementors/reference/#labels). This basically sets some defaults which can be interpreted by devcontainer users like VSCode (unless those properties are overwritten by `.devcontainer.json` file). I thought the container user `ros2` and the mount path for `/home/ros2/ws/src` are very useful defaults to have in there. I could also leave that for a separate PR, but I didn't want to rebuild patches 1.0.1 and 1.0.2 back to back for no reason.

## Review guidelines

The first four revert commits are just plain reverts of #108 in reverse order.

Then review the next three commits in order, which fix AICA metadata, set a specific base image and tag, and add devcontainer metadata accordingly.

## Checklist before merging:


ros2_ws/Dockerfile
```diff
@@ -153,3 +153,7 @@ LABEL org.opencontainers.image.description="AICA base ROS 2 image (includes ros2
 LABEL org.opencontainers.image.version="${VERSION}"
 LABEL org.opencontainers.image.base.name="${BASE_IMAGE}:${BASE_TAG}"
 LABEL tech.aica.image.metadata='{"type":"base/ws","base":{"name":"${BASE_IMAGE}:${BASE_TAG}","version":"${VERSION}"}}'
+LABEL devcontainer.metadata='[{ \
+  "containerUser": "ros2", \
+  "mounts": [ "source=${localWorkspaceFolder}/source,target=/home/ros2/ws/src,type=bind,consistency=cached" ], \
```
domire8, line 158: Weeeell this works well for most repos, but not for dynamic components and hardware collections?

domire8 (commented):
Nice!

ros2_ws/Dockerfile
eeberhard: True, although it's just a default that can always be overridden. I was mostly thinking that the user `ros2` and the workspace path `/home/ros2/ws/src` are two specific things for this image that are worth capturing in the metadata. The main intended public use case will be https://github.com/aica-technology/component-template, and other repos can overwrite as needed. Then again, a devcontainer.json configuration will always be needed to set up the image anyway so I could also drop the `mounts` specification. Let me know what you prefer.

domire8: I lean towards removing it, it feels very specific and tailored for one certain repo/folder setup. Feels almost better if we could have those options directly in the aica-package.toml and its added as part of the development image, but then again we need a devcontainer.json file anyway so I don't know if it's worth it at all.

domire8 (approved):

domire8:
Thanks for tracking down the SHA!

ros2_ws/Dockerfile
```diff
@@ -1,10 +1,156 @@
...
 ARG BASE_TAG=iron
+ARG VERSION=v0.0.0
 LABEL org.opencontainers.image.title="AICA ROS 2 image"
 LABEL org.opencontainers.image.description="AICA base ROS 2 image (includes ros2_control)"
 LABEL org.opencontainers.image.version="${VERSION}"
-LABEL org.opencontainers.image.base.name="docker.io/library/ros:${BASE_TAG}"
-LABEL tech.aica.image.metadata='{"type":"base/ws","base":{"name":"docker.io/library/ros","version":"'${BASE_TAG}'"}}'
\ No newline at end of file
+LABEL org.opencontainers.image.base.name="${BASE_IMAGE}:${BASE_TAG}"
+LABEL tech.aica.image.metadata='{"type":"base/ws","base":{"name":"${BASE_IMAGE}:${BASE_TAG}","version":"${VERSION}"}}'
```
LouisBrunner, line 155: While the value is correct, the formatting isn't. The previous version used `"'${VAR}'"` for a specific reason (`'` strings do not extrapolate variables). This breaks the metadata, see:
```
                "tech.aica.image.metadata": "{\"type\":\"base/ws\",\"base\":{\"name\":\"${BASE_IMAGE}:${BASE_TAG}\",\"version\":\"${VERSION}\"}}"
```

PR #110, opened 2024-05-13
fix: define all build args at the top of the Dockerfile

eeberhard:
## Description

Fix following build failure here:

- https://github.com/aica-technology/docker-images/actions/runs/9061701641/job/24894045916

Build args need to be declared at the top layer, otherwise they might be skipped in intermediate layers and remain undefined. As part of this, I removed the secondary declarations of the build args in the later layer.

## Review guidelines

Estimated Time of Review: 1.1 minutes

## Checklist before merging:

### DIFF
--- a/ros2_ws/Dockerfile
+++ b/ros2_ws/Dockerfile
@@ -1,5 +1,8 @@
 ARG BASE_IMAGE=docker.io/library/ros
 ARG BASE_TAG=iron
+ARG ROS_DISTRO=iron
+ARG VERSION=v0.0.0
+
 FROM ${BASE_IMAGE}:${BASE_TAG} as environment-variables
 ENV DEBIAN_FRONTEND=noninteractive
 ENV PYTHONWARNINGS=ignore:::setuptools.command.install,ignore:::setuptools.command.easy_install,ignore:::pkg_resources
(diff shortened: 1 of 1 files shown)


domire8 (approved):

LouisBrunner:
This is incorrect and actually breaks the metadata:

```
                "devcontainer.metadata": "[{\"containerUser\": \"ros2\"}]",
                "org.opencontainers.image.base.name": ":",
                "org.opencontainers.image.description": "AICA base ROS 2 image (includes ros2_control)",
                "org.opencontainers.image.ref.name": "ubuntu",
                "org.opencontainers.image.title": "AICA ROS 2 image",
                "org.opencontainers.image.version": "",
                "tech.aica.image.metadata": "{\"type\":\"base/ws\",\"base\":{\"name\":\"${BASE_IMAGE}:${BASE_TAG}\",\"version\":\"${VERSION}\"}}"
```

Stages only inherit the `ARG` from the stages they inherit from and top-level `ARG` are seemingly only associated with the first stage, resulting in empty metadata (some is also badly formatted, introduced in https://github.com/aica-technology/docker-images/pull/109#discussion_r1598375768).

Issue #112, opened 2024-05-28
Build jazzy image


domire8:
ROS 2 Jazzy Jalisco is out since a week and we can build a new base image for the coming release.

PR #113, opened 2024-06-02
feat: add stages for jazzy 2.0.0 image

domire8:
## Description

- #112 

This PR solves the issue by adding the necessary stages in the Dockerfile. As always, I just added repos with their most recent tags to the ROS2 workspace until the colcon build would succeed. As mentioned before, the 24.04 image contains a user `ubuntu`, so I had to rename it.

If I understood the CI right, this would now build a jazzy image with the 0.0.1 tag that we could start to try out. After, we'd have to update this tag to 2.0.0 since the 1.X are iron images.

## Review guidelines

## Checklist before merging:


ros2_ws/Dockerfile
```diff
@@ -137,10 +144,45 @@ WORKDIR ${ROS2_WORKSPACE}
...
+
+WORKDIR ${ROS2_WORKSPACE}
+RUN /bin/bash -c "source /opt/ros/${ROS_DISTRO}/setup.bash; colcon build"
+RUN rm -rf ./src
+
+FROM ros2-control-${BASE_TAG} as ros2-control
 
 FROM environment-variables as final
 COPY --from=ros2-control / /
+RUN rm -rf /home/ubuntu
```
bpapaspyros, line 185: A slightly annoying nitpick/request, but would it make more sense to move this command directly after renaming the user?

bpapaspyros (approved):
Dockerfile looks good to me for now

ros2_ws/Dockerfile
domire8: You cannot, it's not there after renaming the user. We do `FROM environment-variables as final` and that directory is actually coming from the `environment-variables` stage where we can't delete it yet since the user is still called ubuntu at this point.

The reason for these 2 lines (183 and 184) is that we want to reduce the amount of layers in the final image. Every `RUN` instruction creates a layer and an image can only have so many layers. That's why it's good practice to reduce layers, one way to do that is to do something like in these two lines, start again from the base image and just copy paste the content of the previous stage into the base. The fact that we then have this directory again is a small side effect of that trick.

bpapaspyros: oh right, I see. Ok, then this sounds good to me. Thanks!

ros2_ws/VERSION.jazzy
```diff
@@ -0,0 +1 @@
+0.0.1
```
eeberhard, line 1: might be nice if we can use pre-release tags like `2.0.0-jazzy.rc.1` to leave room for ordered release candidates without abusing the 0.0.1 tag. I haven't checked deeply how that would propagate with the CI and following package / app builders (@LouisBrunner)

domire8: Currently, the CI couldn't handle this

eeberhard: ok then let's go with the 0.0.1 approach for now

eeberhard (approved):

PR #114, opened 2024-06-04
fix: add ros distro back

domire8:
## Description

I broke something in #113 so this fixes it again. There were a `ROS_DISTRO` and `BASE_TAG` args in the Dockerfile, which I didnt understand, but has a good reason, namely that we provide a specific hash as BASE_TAG to the CI for the correct iron build.

## Review guidelines

## Checklist before merging:


eeberhard (approved):
yes indeed, I should have caught that but I was eager to unblock the jazzy build

PR #115, opened 2024-07-29
fix(jazzy): correct backward_ros tag and release rc2

domire8:
## Description

With tag 1.0.4 and upwards, `backward_ros` is not being detected by ROS so I had to roll back to 1.0.3.

This should be released as v2.0.0-rc1-jazzy.

## Review guidelines

## Checklist before merging:


eeberhard (approved):
Just to check, in the PR description you wrote:
>This should be released as v2.0.0-rc1-jazzy

But you mark rc2 in the PR title and code, so I assume that's a typo in the description.

Also, I guess the existing `v2.0.0-rc1-jazzy` was tagged and pushed manually since #113 was merged with 0.0.1 as the version tag for jazzy? 

Overall looks good to me!

domire8:
> But you mark rc2 in the PR title and code, so I assume that's a typo in the description.
Typo yes.

> Also, I guess the existing `v2.0.0-rc1-jazzy` was tagged and pushed manually since #113 was merged with 0.0.1 as the version tag for jazzy?
That's correct yeah, the action was triggered from a branch

Issue #118, opened 2024-08-29
Create generic RViz configuration file


domire8:
Add a default RViz config in the home folder of the user such that the launcher can use it to open RViz with it

PR #119, opened 2024-09-09
feat: add default rviz config

domire8:
## Description

- #118 

This PR solves the issue by adding a configuration with TF and a robot model. The robot model is expected under `hardware/robot_description` which should be correct once https://github.com/aica-technology/frontend/issues/1053 is through.

Additionally, I updated to the most recent tags of ros2 control and its dependencies, so we'll have to see what updates we need to make to account for that

## Review guidelines

## Checklist before merging:


bpapaspyros (approved):

domire8:
Fairly happy with that, it seems that the required changes downstream are minimal and I've already tested a robot with a mock interface with this new base image

eeberhard (approved):
I would have preferred these two changes to be separate (rviz and ros2_control) but let's go for it 👍

domire8:
> I would have preferred these two changes to be separate (rviz and ros2_control) but let's go for it 👍

You're right, especially if we want to revert stuff this would be cleaner

PR #122, opened 2024-10-30
fix: don't show tf axes by default in rviz

domire8:
## Description

Just a minor thing that was bugging me, the axes should not necessarily be on by default, just the robot model alone is better.
We can then use the 2.0.1 version next time we release a new core.

## Review guidelines

Estimated Time of Review: 1.1 minutes

## Checklist before merging:


bpapaspyros (approved):
fair enough :+1:

Issue #124, opened 2025-03-10
Generalize the generate_compile_commands for ros2_ws


bpapaspyros:
We have been updating our repos to improve the devcontainer experience, which means we are now testing out/shifting to using clangd for linting and code completion. This requires a `compile_commands.json` to be generated which we have so far been doing with repo-specific scripts.

Since this is going to be a common step for different repos, we should consider including a generic version of this script in our `ros2_ws` image

PR #125, opened 2025-03-10
feat: add generic export compile commands script for ROS workspaces

bpapaspyros:
## Description

- #124 

This PR solves the issue by adding a script to the repo and copying it as part of the Dockerfile to `/tmp`

## Review guidelines

## Checklist before merging:


ros2_ws/Dockerfile
```diff
@@ -190,6 +190,8 @@ RUN rm -rf /home/ubuntu
 USER ${USER}
 WORKDIR ${ROS2_WORKSPACE}
 
+COPY ./scripts/generate_compile_commands.sh /tmp/generate_compile_commands
```
domire8, line 193: Could you do that in line 94?

ros2_ws/scripts/generate_compile_commands.sh
domire8, line 1: also feel that this could go into the config folder instead of scripts

ros2_ws/scripts/generate_compile_commands.sh
```diff
@@ -0,0 +1,60 @@
...
+COLCON_PID=$!
+TIMER=0
+touch compile_commands.json
+for PACKAGE in $PACKAGES; do
+  while ! [ -s build/$PACKAGE/compile_commands.json ] && [ $TIMER -lt $TIMEOUT ] ; do
+    echo "waiting for package $PACKAGE ($TIMER)"
+    sleep 1
+    TIMER=$((TIMER + 1))
+  done
+  jq -s 'add' compile_commands.json build/$PACKAGE/compile_commands.json > tmp && mv tmp compile_commands.json
```
domire8, line 57: This fails if the timeout elapsed before everything is built, might be nice to catch that and return an error such that this becomes visible?

bpapaspyros: I added just an echo message but didn't fail the entire script. I figure there might still be reason to continue with the other packages and I guess linting for that one would just not happen

domire8 (approved):

PR #126, opened 2025-03-11
feat: add print encoded topic alias in ros2_ws

bpapaspyros:
## Description

- #127 

This PR solves the issue by adding an alias for the print encoded topic node and also adding autocomplete to get ros topics.

## Review guidelines

Estimated Time of Review: 4 minutes

## Checklist before merging:


bpapaspyros:
@domire8 do we cherry-pick and release 2.0.3 again or how would you like to handle that?

ros2_ws/config/general_aliases.sh
LouisBrunner, line 1: Can't we run this file directly inside Docker? It's clearer and avoid a bunch of extra layers.

domire8: What do you mean exactly with "directly inside Docker"?

LouisBrunner:
```Dockerfile
RUN <<EOC
echo "alias print_encoded_topic='export ROS2_WORKSPACE; ros2 run aica_core_components print_encoded_topic'" >> "${HOME}"/.bash_aliases

cat << 'EOF' >> "${HOME}/.bash_aliases"
_print_encoded_topic_completion() {
  local cur_word
  cur_word="${COMP_WORDS[COMP_CWORD]}"
  local topics
  topics=$(ros2 topic list 2>/dev/null)
  COMPREPLY=( $(compgen -W "$topics" -- "$cur_word") )
}
complete -F _print_encoded_topic_completion print_encoded_topic
EOF
EOC
```

bpapaspyros: yup we could do that, I just figured that we might want to add things later on and a standalone script would result in a cleaner Dockerfile. 

no strong opinion on this though

bpapaspyros:
also fyi, I tested this version:

```shell
_print_encoded_topic_completion() {
  local cur_word filtered topics
  cur_word="${COMP_WORDS[COMP_CWORD]}"
  topics=$(ros2 topic list 2>/dev/null)
  filtered=""
  for topic in $topics; do
    if ros2 topic info "$topic" 2>/dev/null | grep -q 'Type: modulo_interfaces/msg/EncodedState'; then
      filtered="$filtered $topic"
    fi
  done
  COMPREPLY=( $(compgen -W "$filtered" -- "$cur_word") )
}
```

which would only autocomplete encoded topics. This is however very slow and I tested with just a six-axis robot and state broadcaster, so I'd probably stay away from it (still leaving it here for archiving purposes)

bpapaspyros:
moved everything in the Dockerfile. Thanks @domire8 with the -t option I can now easily filter out the encoded topics. This is pretty fast now

LouisBrunner (approved):
🚀

domire8:
> Do we cherry-pick and release 2.0.3 again or how would you like to handle that? (if we agree to actually merge)

We merge first, then cherry-pick this commit on top of v2.0.2-jazzy tag, then push this as v2.0.3-jazzy tag and use the branch to manually trigger the build push workflow from that branch with the force option

domire8:
Let's move these lines into the Dockerfile of our core and close this PR without merge

bpapaspyros:
Closing as this will be handled as part of our core build, not the `ros2_ws`

PR #131, opened 2025-04-23
feat: install json validation tool

domire8:
## Description

This PR adds a utility program written and built with Go to validate our component/controller descriptions in JSON. It was necessary to do that because most available tools for json validation don't parse the `$schema` tag inside the json instance. 
The `validate_json` tool can then be used in a Cmake macro to wrap the validation and installation of descriptions.

Once this is through, ROS packages can then just do
```
list(APPEND CMAKE_MODULE_PATH "/home/ros2/.cmake")
include(InstallAicaDescriptions)
install_aica_descriptions(./component_descriptions ${CMAKE_INSTALL_PREFIX})
```
and you'll get feedback like
```
Starting >>> template_component_package
--- stderr: template_component_package                         
Validating and installing descriptions
CMake Error at /home/ros2/.cmake/InstallAicaDescriptions.cmake:12 (message):
  Failed to validate file
  /home/ros2/ws/src/template_component_package/./component_descriptions/template_component_package_cpp_component.json:Error
  unmarshalling JSON instance: invalid character '}' looking for beginning of
  object key string

Call Stack (most recent call first):
  CMakeLists.txt:47 (install_aica_descriptions)
```
if the JSON instance is not valid

## Review guidelines

## Checklist before merging:

- [N/A ] Confirm that the relevant changelog(s) are up-to-date in case of any user-facing changes


ros2_ws/config/validate_json.go
```diff
@@ -0,0 +1,103 @@
...
+		fmt.Printf("Error reading JSON instance file '%s': %v\n", instanceFile, err)
+        os.Exit(1)
+		return
+	}
+
+	var instance map[string]interface{}
+	err = json.Unmarshal(instanceData, &instance)
+	if err != nil {
+		fmt.Printf("Error unmarshalling JSON instance: %v\n", err)
+        os.Exit(1)
```
bpapaspyros, line 61: nit: should these `os.Exit` be indented more?

bpapaspyros (approved):
looks good to me me. Especially cmake/docker, I leave it the others for go code.

ros2_ws/config/InstallAicaDescriptions.cmake
```diff
@@ -0,0 +1,16 @@
+macro(install_aica_descriptions)
```
LouisBrunner, line 1: Why not

```suggestion
macro(install_aica_descriptions BASE_FOLDER OUT_FOLDER)
```

and avoid dealing with `ARGX`?

ros2_ws/Dockerfile
```diff
@@ -92,6 +108,8 @@ WORKDIR ${HOME}
 RUN wget https://raw.githubusercontent.com/aica-technology/.github/v0.9.0/guidelines/.clang-format
 COPY ./config/config.rviz config.rviz
 COPY --chown=${USER}:${USER} ./config/generate_compile_commands.sh /usr/local/bin/generate_compile_commands
+RUN mkdir -p ${HOME}/.cmake
+COPY --chown=${USER}:${USER} ./config/InstallAicaDescriptions.cmake .cmake/
```
LouisBrunner, line 112: Is `~/.cmake` a standard folder which is easy to import? I thought `/usr/share/cmake` was the standard place.

ros2_ws/Dockerfile
```diff
@@ -56,6 +56,22 @@ RUN ( \
...
+ARG GOVERSION=1.24.2
+
+RUN wget https://go.dev/dl/go${GOVERSION}.linux-${TARGETPLATFORM#*/}.tar.gz -O /tmp/go.tar.gz && \
+  sudo tar -C /usr/local -xzf /tmp/go.tar.gz && \
+  rm /tmp/go.tar.gz
+ENV PATH="${PATH}:/usr/local/go/bin"
+WORKDIR /go
+COPY config/validate_json.go .
+RUN --mount=type=cache,target=/go/pkg/mod \
+  go mod init validate_json.go && go mod tidy
```
LouisBrunner, line 70: Shouldn't run `tidy` in the Docker image as it might mutate the `go.mod`/`go.sum` and thus install things that aren't pinned in source. Instead we should have the `go.mod` in source control.
Maybe just create a `validate_json` subfolder in `config` so you can group all the files easily? (you can then also avoid the `-o validate_json` as `go build` derive the output name from the name of the parent folder of the `package main`, another funky quirk).

ros2_ws/config/InstallAicaDescriptions.cmake
domire8: Good idea, didn't know it was possible

ros2_ws/Dockerfile
```diff
@@ -92,6 +106,8 @@ WORKDIR ${HOME}
 RUN wget https://raw.githubusercontent.com/aica-technology/.github/v0.9.0/guidelines/.clang-format
 COPY ./config/config.rviz config.rviz
 COPY --chown=${USER}:${USER} ./config/generate_compile_commands.sh /usr/local/bin/generate_compile_commands
+COPY --chown=${USER}:${USER} ./config/InstallAicaDescriptions.cmake /tmp
+RUN sudo mv /tmp/InstallAicaDescriptions.cmake /usr/share/cmake-$(cmake --version | awk '/cmake version/ {split($3, v, "."); print v[1]"."v[2]}')/Modules
```
domire8, line 110: This is maybe a bit brittle but it works

domire8 (commented):
Thanks @LouisBrunner for the helpful tips, it looks much better now

LouisBrunner (approved):
Solid v1

eeberhard (approved):
Nice!

PR #138, opened 2025-06-03
fix: ros apt keys (temporary part 2)

domire8:
## Description

- #135 

An oversight in the prevoius PR sorry, the final stage needs the two files removed again.

## Review guidelines

## Checklist before merging:


bpapaspyros (approved):
fair. Seems good to me, but haven't tested

PR #140, opened 2025-06-13
feat: add camera calibrator image

bpapaspyros:
## Description

- #141 

This PR solves the issue by adding the camera calibration package in our image.

The code in itself will not run without additional dependencies. I added those to the Dockerfile under `camera_calibration`.

You can build by:

`./build.sh`

and run with `./run.sh`

The calibration result and sample images are stored in `/tmp` so we'll have to mount that too.

## Review guidelines

## Checklist before merging:

### Blocked by:
- #142

### DIFF
--- a/camera_calibration/run.sh
+++ b/camera_calibration/run.sh
@@ -1,5 +1,9 @@
+#!/bin/bash
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
+
 VERSION=$(cat VERSION)
-OUTPUT_DIR=calibration
+OUTPUT_DIR="${SCRIPT_DIR}"/calibration
 CALIB_WIDTH=7
 CALIB_HEIGHT=9
 CALIB_SQUARE=0.015


bpapaspyros:
Not sure what would be best here. I initially suggested we keep some dependencies in our ros2_ws, but perhaps we can just release a single image with a `-camera-calibration` tag that has everything in it. In theory we don't even need to maintain it much if its sole purpose is to run the calibration software

domire8 (commented):
I'm still not super happy with this. Either we start with a ros2-ws image and just add calibration or we do a separate Dockerfile in a separate directory because it doesn't make sense to have all the ros2 control dependencies and whatever else we have if this is just meant to run the calibration

bpapaspyros:
> I'm still not super happy with this. Either we start with a ros2-ws image and just add calibration or we do a separate Dockerfile in a separate directory because it doesn't make sense to have all the ros2 control dependencies and whatever else we have if this is just meant to run the calibration

ah I see, my bad, I didn't quite understand the previous comment then. Let me take another look at it then

bpapaspyros:
it's still a pretty big image for the calibrator alone (2.5 GB) but significantly less than the one that has our ros2 ws + calibrator (3.8 GB)

domire8:
And what if you use the ros:jazzy directly? It is 875MB. I would also say that if you put it in it's own directory we can even put it on main. It can live in parallel to the ros2-ws. We also dont need the iron version because we just want to extract the calibration

bpapaspyros:
> And what if you use the ros:jazzy directly? It is 875MB. I would also say that if you put it in it's own directory we can even put it on main. It can live in parallel to the ros2-ws. 

Ok, sure, I'll give it a try. I thought we'll just keep it on the branch but it wouldn't hurt indeed.

> We also dont need the iron version because we just want to extract the calibration
true, yes

ros2_ws/Dockerfile
```diff
@@ -17,14 +17,6 @@ ENV COLCON_WORKSPACE=${ROS2_WORKSPACE}
 
 FROM environment-variables AS base-dependencies
 
-# FIXME: remove after new ros images are released
```
bpapaspyros, line 20: we can remove this separately, but the image won't build otherwise

camera_calibration/config/sshd_entrypoint.sh
domire8, line 1: File not needed

ros2_ws/Dockerfile
domire8, line 1: Revert the changes here please

domire8 (commented):
Nice yes looks good. When I run, I also get the GUI now but it's black. Do you know what that could be? it should show the integrated cam no?

PR #142, opened 2025-06-18
fix: remove temporary ROS keyring fix

bpapaspyros:
## Description

- n/a

Now that ROS updated its keys, we can remove the temporary fix

## Review guidelines

Estimated Time of Review: 1 minutes

## Checklist before merging:


domire8 (approved):
I allowed myself to force push directly here, would prefer to properly revert those commits. You can rebase and merge :+1:

bpapaspyros:
thanks!

PR #147, opened 2025-07-10
feat: create ros2-ws with nvidia base (cuda, pytorch, etc installed)

bpapaspyros:
## Description

- #146 

Ok, this is an experimental PR where I wanted to see if it would make more sense that we have a dedicated base with Nvidia dependencies. This could give us a ros2-ws that has all cuda dependencies preinstalled (this is the pytorch image, so pytorch too which can be tricky) and on top of that we replicate our ros2-ws. 

I've now tested this both on humble and jazzy and it doesn't seem to create any unexpected issues, albeit we still need to test this.

Notable points:
- during build this image inflates a lot (upwards to 120gb)
- final image is ~29gb
- leftover build overlays are ~60gb

I am not sure if we can tidy up the layering for this, but let's worry about this if/when we agree this image makes sense altogether (fyi @LouisBrunner)

Other considerations:
- We might still want to have ros2-ws image with some additional machine learning packages pre-packaged, but this one would be an overkill coming in at 29GB (the fact that it is Nvidia-related is not an issue for non-Nvidia enabled devices, but means no GPU will be used by pytorch if you don't have one and practically you have a lot of useless packages). On the other hand, if we are not in need of cuda installation and pytorch, we can already handle ML dependencies, so there is probably no need for another image for non-Nvidia ML ros2-ws.

**This is WIP so I will potentially spam you with some notification since I tagged you, apologies**

## Review guidelines

## Checklist before merging:


ros2_ws/Dockerfile.nvidia
```diff
@@ -0,0 +1,285 @@
+ARG NVIDIA_BASE_IMAGE=nvcr.io/nvidia/pytorch
+ARG NVIDIA_BASE_TAG=25.06-py3
```
bpapaspyros, line 2: this is the latest version ubuntu 24.04 and python 3.12 with very recent cuda versions as well. I think 24.12 might be a bit more fitting for 22.04 devices. I will have to verify, but ultimately taking a small step back in this version is not going to be harmful.

For our humble images we'd have to go further back to ensure the necessary python and cuda version I believe

bpapaspyros:
Compared to injecting cuda dependencies with package builder, this generates a bigger image, but it also carries a lot more nvidia-related packages (I can report on where these GB are distributed, but it's mostly nvidia things).

This may be good for future-proofing, but 29GB is intimidating. We could potentially clean up a bit if we go in the direction of having such a workspace

bpapaspyros:
fyi @yrh012 let's see where the discussion takes us and perhaps we'll need to add some common packages. I'd avoid doing so if not very critical (size image is very big, bigger than current images but at least comparable)

ros2_ws/Dockerfile.nvidia
```diff
@@ -0,0 +1,304 @@
...
+# # RUN git clone -b 1.0.2 --depth 1 https://github.com/pal-robotics/backward_ros.git
+# # RUN git clone -b 3.1.2 --depth 1 https://github.com/ros/diagnostics.git
+# # RUN git clone -b 0.3.3 --depth 1 https://github.com/PickNikRobotics/generate_parameter_library.git
+# # RUN git clone -b 0.2.2 --depth 1 https://github.com/PickNikRobotics/RSL.git
+# # RUN git clone -b 1.0.2 --depth 1 https://github.com/PickNikRobotics/cpp_polyfills.git
+# # RUN git clone -b 0.1.0 --depth 1 https://github.com/ros-controls/kinematics_interface.git
+
+WORKDIR ${ROS2_WORKSPACE}
+RUN /bin/bash -c "source /opt/ros/${ROS_DISTRO}/setup.bash; colcon build"
+# RUN rm -rf ./src
```
bpapaspyros, line 214:
```suggestion
RUN rm -rf ./src
```

bpapaspyros:
ok, after some internal discussions we've decided to address this differently. I am closing this PR in favour of smaller ones coming up where I'll tag you accordingly

PR #150, opened 2025-07-21
feat: first cuda and ML library images

bpapaspyros:
## Description

- #149  and #148 

I considered splitting the PR because now I need to have 2 parent issues, but they are very correlated at this point and easier to test, so bear with me. Perhaps we can split once the testing is done, but I see little added value there.

Two new images:
- Cuda toolkit (defaults to 12.6 libraries)
- ML toolkit (installs some common Python and C++ dependencies and depending on parameters has cpu and gpu versions)

ToDO:
- we need a naming convention that differentiates between cpu and gpu (currently not handled; simply append the keyword seems the easiest)
- for Cuda, I think we should also be appending the version, e.g., `cu126` in this default case
- I am putting some more technical notes as code comments

This is still WIP while I am validating across devices and applications, but feel free to share initial feedback if you have any.

## Review guidelines

Estimated Time of Review: 20 minutes

## Checklist before merging:


toolkits/build.sh
```diff
@@ -0,0 +1,88 @@
...
+
+  BUILD_FLAGS+=(--build-arg=UBUNTU_VERSION=${UBUNTU_VERSION})
+  BUILD_FLAGS+=(--build-arg=PYTHON_VERSION=${PYTHON_VERSION})
+  BUILD_FLAGS+=(--build-arg=TORCH_VARIANT=${TORCH_VARIANT})
+  BUILD_FLAGS+=(--target ${TARGET})
+fi
+
+BUILD_FLAGS+=(--build-arg=TRT_IMAGE_TAG=${TRT_IMAGE_TAG})
+BUILD_FLAGS+=(--build-arg=VERSION=${VERSION})
+docker buildx build -t "${IMAGE_NAME}":v"${VERSION}" "${BUILD_FLAGS[@]}" .
```
bpapaspyros, line 88: ToDo: I'll append gpu/cpu/cu versions eventually, but feel free to share your preferred naming convention

toolkits/Dockerfile.ml
```diff
@@ -0,0 +1,221 @@
...
+    python3-pybind11 \
+    python3-dev \
+    libssl-dev \
+    libffi-dev \
+    flatbuffers-compiler \
+    libeigen3-dev \
+    tar \
+  && rm -rf /var/lib/apt/lists/* \
+  && update-ca-certificates \
+  && mkdir -p ${CPP_DEPS}/lib ${CPP_DEPS}/include
```
bpapaspyros, line 108: I could probably partially copy apt deps from another stage, but at this version of the image I don't think it will have a big impact

toolkits/Dockerfile.cuda
```diff
@@ -0,0 +1,81 @@
...
+FROM scratch AS final-builder
+
+ARG CUDA_DEPS
+WORKDIR /
+
+COPY --from=builder ${CUDA_DEPS} /deps
+
+# ! this is me exploiting a bug we have (it's ugly, I know), we need a much more consistent way to handle this
+# ! but helps for debugging and development for now...
+COPY --from=builder ${CUDA_DEPS}/cuda_env.sh /ws/install/aica_event_engine/apt/opt/ros/jazzy/setup.bash
```
bpapaspyros, line 62: todo: @LouisBrunner this is the point I brought up. Which dir should I target? I tested profile.d with no success

LouisBrunner: Yeah, that's tricky. Without adding our own system (which wouldn't be a big deal but it would need to be done everywhere) we could still leverage the colcon/ament hook system? If we created a dsv file into `/home/ros2/ros2_ws/install/ros2_control` which sources this one then it should work in all use cases?

bpapaspyros: Unfortunately colcon somehow hardcodes which dsvs to source at runtime, so just dropping a dsv in the directory doesn't work. (I tried injecting it in one of our own packages with hooks)

There is another alternative. We can create a ros package as part of ros2-ws and create a package that only has a single hook for a script that source everything in the same package. Then, we could always target that package for hooks, but honestly:
1. that seems a bit weird to target a ros package
2. if we are changing the ros2-ws we might as well just have a /environment_vars directory be sourced automatically instead of a ROS package

LouisBrunner: I would be happy with an empty ROS package which just source the right environment. I agree it's weird but ROS already has a system to do sourcing and we source the workspace everywhere (dev-mode, prod, etc, etc) so building a parallel system seems a bit wasteful.

bpapaspyros: ok, I'll give that a go. For now this can be isolated in one of the toolkit images in any case. We can come back to this topic if we start seeing more images needing this mechanism

bpapaspyros: done, works as expected, I think it's good for now

bpapaspyros:
I am now opening this for review. There is more testing to be done, but this is now verified on the Jetson and we only need to verify on `amd64` and get some time testing this to check for runtime issues/missing packages. So any feedback is welcome.

toolkits/Dockerfile.cuda
```diff
@@ -0,0 +1,150 @@
...
+  "hooks": ["share/environment_variables/environment/00-environment-variables.sh"]
+}
+EOF
+
+WORKDIR $ROS_WS
+RUN bash -c "source /opt/ros/${ROS_DISTRO}/setup.bash && \
+  colcon build"
+RUN cp -r $ROS_WS/install/environment_variables /environment_variables
+
+FROM scratch AS final-builder
```
domire8, line 121: Why do we have two stages that go from `scratch` and then yet another `env-vars` which seems to be the final stage?

toolkits/Dockerfile.ml
```diff
@@ -0,0 +1,268 @@
...
+ARG PY_DEPS=/tmp/py_deps
+ARG PYTHON_VERSION=3.12
+ARG UBUNTU_VERSION=24.04
+
+ARG TRT_IMAGE_TAG=24.12-py3
+ARG TENSORRT_IMAGE=nvcr.io/nvidia/tensorrt
+
+FROM python:${PYTHON_VERSION}-slim AS python-builder
+
+ARG TORCH_VARIANT=cpu
```
domire8, line 11: Don't we need to define those up top in order to be able to change them?

toolkits/README.md
```diff
@@ -0,0 +1,124 @@
...
+> specialized Linux for Tegra (L4T) images, which you can browse
+> [here](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/l4t-tensorrt). AICA's Jetson images are also based on L4T
+> images.
+> 
+> You may also need to install custom versions of PyTorch that are built for Tegra (or compile wheels from source). You
+> may refer to the this [Nvidia](https://forums.developer.nvidia.com/t/pytorch-for-jetson/72048) forum post.
+
+## Using registry images
+
+To use registry images you may simply add the following line(s) in your `aica-application.toml`:
```
domire8, line 53: That yes, but also how to use it with package builder under dependencies

toolkits/Dockerfile.ml
bpapaspyros: not really, as long as it's defined under the correct stage. But it is a question of visibility for someone that wants to read the Dockerfile, so I'm not against putting everything up top too

toolkits/Dockerfile.ml
```diff
@@ -0,0 +1,268 @@
...
+RUN apt-get update \
+ && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+    ca-certificates \
+    wget \
+    g++
+
+RUN pip install --upgrade pip \
+ && mkdir -p ${PY_DEPS}
+
+RUN if [ "${TORCH_VARIANT}" = "jetson" ]; then \
```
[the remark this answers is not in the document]
bpapaspyros: hard to say to be fair. Perhaps it's more like "any tegra", but even then I don't have the evidence to support that. It does look like it though

domire8 (commented):
Looks good with me, one commend still pending above

toolkits/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+[build]
+type = "ros"
+image = "v2.0.5-jazzy"
+
+[build.dependencies]
+"@aica/foss/control-libraries" = "v9.2.0"
+"@aica/foss/modulo" = "v5.2.0"
+
+# along with the above core libraries, you can also include these ancillary images
```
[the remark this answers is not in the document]
bpapaspyros: sorry, that's my phd days speaking and a bit of brain fog after spending time with CUDA :smile: . I'll update it

toolkits/Dockerfile.cuda
```diff
@@ -0,0 +1,134 @@
...
+ARG CUDA_DEPS
+COPY --from=builder ${CUDA_DEPS} /
+
+ARG VERSION=0.0.0
+LABEL org.opencontainers.image.title="AICA CUDA libraries"
+LABEL org.opencontainers.image.description="AICA CUDA libraries"
+LABEL org.opencontainers.image.version="${VERSION}"
+LABEL tech.aica.image.metadata='{"type":"lib"}'
+
+FROM barebones AS env-vars
```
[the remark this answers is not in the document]
bpapaspyros: I think typically we are going to be using the one with env vars (because we need the variables for our packaged core). But since the whole sourcing procedure is done through a ROS package, I figured perhaps we should also support a more lean version (although that package is just text files, not a big concern for size)

I don't mind if we directly go into a single final stage with the ROS package (technically, even if someone doesn't need it for whatever reason, it doesn't hurt having it)

toolkits/Dockerfile.cuda
```diff
@@ -0,0 +1,131 @@
...
+  "hooks": ["share/environment_variables/environment/00-environment-variables.sh"]
+}
+EOF
+
+WORKDIR $ROS_WS
+RUN bash -c "source /opt/ros/${ROS_DISTRO}/setup.bash && \
+  colcon build"
+RUN cp -r $ROS_WS/install/environment_variables /environment_variables
+
+FROM scratch
```
domire8, line 120: Nitpick but would prefer that each stage has a name, so maybe
```suggestion
FROM scratch AS final
```

bpapaspyros:
ok, overall this is looking good on amd64, but I need to figure out some python package  version issues.

I am turning this back into draft mode until I have resolved them

domire8:
> I think this is a good candidate for an RC version once the reviews are done

Sounds good to me too. Do you want to add the `-rc1` directly in the version files or what is your plan? Do we need to integrate it in the CI or do we wait with that a bit?

bpapaspyros:
> > I think this is a good candidate for an RC version once the reviews are done
> 
> Sounds good to me too. Do you want to add the `-rc1` directly in the version files or what is your plan? Do we need to integrate it in the CI or do we wait with that a bit?

I can add the rc1, it's then just a question of if we like the current tags I have (doesn't have to be the final one). I think it would be nice to get those in our registry as RCs so that testers don't have to go through the build process, so I guess that would have to be in the CI, yes

bpapaspyros:
ok, I added support for RC flags (needed a bit of coding to get the RC tag to always be at the end). Our new tags will be for example:

```shell
0.1.0-24.12-py3                      # cuda
0.1.0-gpu-24.12-py3-rc0001 # ml
```

toolkits/Dockerfile.ml
```diff
@@ -0,0 +1,268 @@
...
+      -c /tmp/constraints.txt \
+      -r /tmp/requirements.txt
+
+FROM scratch AS cpu
+
+ARG PYTHON_VERSION
+ARG CPP_DEPS
+ARG PY_DEPS
+
+COPY --from=apt-fetch /deps/lib /usr/local/lib
```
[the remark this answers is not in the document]
LouisBrunner: `/usr/local` isn't usually included by default when looking for libraries, includes, etc. So I would say if you expect the things inside to just plug'n'play you should probably just use `/usr`.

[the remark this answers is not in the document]
bpapaspyros: I would still leave Cuda deps under /usr/local since this is it's default path, but I'll move the ML ones

toolkits/Dockerfile.ml
```diff
@@ -0,0 +1,268 @@
...
+LABEL org.opencontainers.image.version="${VERSION}"
+LABEL tech.aica.image.metadata='{"type":"lib"}'
+
+FROM scratch AS gpu
+
+ARG PYTHON_VERSION
+ARG CPP_DEPS
+ARG PY_DEPS
+
+COPY --from=apt-fetch /deps/lib /usr/local/lib
```
bpapaspyros, line 257:
```suggestion
COPY --from=apt-fetch /deps/lib /usr/lib
```

PR #153, opened 2025-08-07
feat: add toolkit image workflows

bpapaspyros:
## Description

- #151 

I am adding some new workflows. It's untested and with many todos before I can officially mark this as ready, but I'd appreciate the feedback (especially on version handling here).

side note, I'm figuring out how to parametrize this nicely, because so far there are a lot of build args that we need to set.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/toolkits/README.md
+++ b/toolkits/README.md
@@ -63,8 +63,8 @@ image = "v4.4.2"
 
 [packages]
 <Other packages you wish to include>
-"@aica/toolkits/cuda-toolkit" = "v0.1.0-cuda24.12-rc0001"
-"@aica/toolkits/ml-toolkit" = "v0.1.0-gpu24.12-rc0001"
+"@aica/toolkits/cuda" = "v0.1.0-cuda24.12-rc0001"
+"@aica/toolkits/ml" = "v0.1.0-gpu24.12-rc0001"
 ```
 
 If your custom component is depending on some of the included packages (e.g., CUDA libraries, torch, ...), note that you
@@ -88,8 +88,8 @@ image = "v2.0.5-jazzy"
 
 # along with the above core libraries, you can also include the toolkit
 # images for easy access to CUDA and ML libraries
-"@aica/toolkits/cuda-toolkit" = "v0.1.0-cuda24.12-rc0001"
-"@aica/toolkits/ml-toolkit" = "v0.1.0-gpu24.12-rc0001"
+"@aica/toolkits/cuda" = "v0.1.0-cuda24.12-rc0001"
+"@aica/toolkits/ml" = "v0.1.0-gpu24.12-rc0001"
 
 [build.packages.my_custom_component]
 source = "./source/my_custom_component"
@@ -111,7 +111,7 @@ For example, if you specify:
   --target env-vars
 ```
 
-you will obtain a `ghcr.io/aica-technology/cuda-toolkit:vX.Y.Z-cpu24.04-rc0001` image (refer
+you will obtain a `ghcr.io/aica-technology/toolkits/cuda:vX.Y.Z-cpu24.04-rc0001` image (refer
 [here]((https://docs.nvidia.com/deeplearning/frameworks/container-release-notes/index.html)) for information on CUDA and
 TensorRT versions).
 
@@ -123,7 +123,7 @@ Similarly, you can do:
   --target gpu
 ```
 
-to obtain `ghcr.io/aica-technology/cuda-toolkit:vX.Y.Z-gpu24.12-rc0001`.
+to obtain `ghcr.io/aica-technology/toolkits/cuda:vX.Y.Z-gpu24.12-rc0001`.
 
 ### Example 2
 
@@ -153,5 +153,5 @@ the registries you would have to specify quite a few options and run the build s
 Those are only few of the available arguments you can provide. Notice how here we specify L4T images as our base. These
 commands will generate the following images:
 
-- `ghcr.io/aica-technology/cuda-toolkit:vX.Y.Z-l4tr8.6.2-rc0001`
-- `ghcr.io/aica-technology/ml-toolkit:vX.Y.Z-l4tr8.6.2-rc0001`
+- `ghcr.io/aica-technology/toolkits/cuda:vX.Y.Z-l4t8.6.2-rc0001`
+- `ghcr.io/aica-technology/toolkits/ml:vX.Y.Z-l4t8.6.2-rc0001`


.github/workflows/toolkit-build-push.yml
```diff
@@ -0,0 +1,132 @@
...
+    outputs:
+      image_name: ${{ steps.ensure-image.outputs.image_name }}
+      image_tags: ${{ steps.tags.outputs.image_tags }}
+      build_flags: ${{ steps.tags.outputs.build_flags }}
+      git_tag: ${{ steps.tags.outputs.git_tag }}
+    steps:
+      - uses: aica-technology/.github/.github/actions/ghcr-ensure-prefix@v0.6.0
+        id: ensure-image
+        with:
+          image_name: aica-technology/${{ inputs.variant }}-toolkit # TODO: that's my original naming, but perhaps toolkit-${{ inputs.variant }} would be better?
```
LouisBrunner, line 65: I agree with the comment, maybe even `aica-technology/toolkits/${{ inputs.variant }}` if that works?

.github/workflows/build-push.yml
```diff
@@ -15,12 +15,87 @@ jobs:
...
+        --build-arg TENSORRT_IMAGE=nvcr.io/nvidia/tensorrt
+        --build-arg TRT_IMAGE_TAG=24.12-py3
+        --build-arg VERSION=0.1.0-gpu-24.12-py3-rc0001 # TODO: remove those from here and handle them in the build
+    secrets:
+      token: ${{ secrets.GITHUB_TOKEN }}
+
+  build-ml-toolkit-for-cpu:
+    uses: ./.github/workflows/toolkit-build-push.yml
+    with:
+      variant: ml
```
LouisBrunner, line 47: Should this be `ml-cpu` and the other `ml-gpu`? How do these tags not conflict?

bpapaspyros: ml just gives the job variant. GPU or CPU is determined by the `TARGET` and appended to the version. 

In general, the versioning scheme here is something that allowed me to test (and made sense to me at the time), but I'm open to changing how things are done and named

LouisBrunner: I can't see in the workflow where `image_tags` includes anything else than the variant? To me it looks like apart from these 2 tags all the other would be the same for cpu/gpu?

bpapaspyros: ah, sure, yes. The intended naming convention is the one on the build script. I didn't do the full naming in the workflow (hence the TODO) until we agree on a naming convention. 

I think the images can now be:

e.g. for cuda

ghcr.io/aica-technology/toolkits/cuda:vX.Y.Z-<TARGET>-<TENSORRT version>

bpapaspyros: you are right, for now, because jetson needs to pull from a completely different TENSORRT base it kind of natively has a very different TENSORRT_VERSION. 

I don't think that will change anytime soon for nvdia, but we might indeed want to append something with jetson. Two examples:


ghcr.io/aica-technology/toolkits/ml:v0.1.0-gpu-r8.6.2-devel # jetson

vs

ghcr.io/aica-technology/toolkits/ml:v0.1.0-gpu-24.12-py3 # non-jetson


what do you think?

bpapaspyros: (we can also probably remove the -devel and -py3 since we are at it)

bpapaspyros: each of those numbers means:

- different python
- different cuda compatibiilty

So Cuda and ML need the same base suffix at all times and should be checked. Then our version just signifies our changes on top of the TensorRT base, and we can check as we are already doing in the current images

LouisBrunner: I think it would be nice to have `-jetson` like `-gpu-jetson` just to avoid any confusion. Otherwise you have to know which technology is in which version range which is not the most explicit.

Everything else LGTM 😄

bpapaspyros: ok, reworked the versions. We'll now have

CUDA:

- 0.1.0-jetson-r8.6.2-rc0001
- 0.1.0-24.12-rc0001

ML:

- 0.1.0-jetson-gpu-r8.6.2-rc0001
- 0.1.0-gpu-24.12-rc0001
- 0.1.0-cpu-24.12-rc0001


A bit lengthy for the RCs, but that's ok-ish I hope

bpapaspyros: (we can drop gpu for jetson, I think it's pretty self-explanatory)

bpapaspyros: > LGTM. Is that with the TensorRT version as well?
> 
> > 0.1.0-24.12-rc0001
> 
> Should this have an extra tag between the two versions?

I can do `0.1.0-trt24.12-rc0001`

LouisBrunner: Does that make sense/work for you:

CUDA:

 * 0.1.0-jetson8.6.2-rc0001
 * 0.1.0-trt24.12-rc0001 (this one works for CPU & GPU?)

ML:

 * 0.1.0-jetson8.6.2-rc0001
 * 0.1.0-gpu-trt24.12-rc0001
 * 0.1.0-cpu-trt24.12-rc0001

bpapaspyros: Thanks again for the feedback, and sorry for the back n' forth, it's always hard to find good but short names

> Does that make sense/work for you:
> 
> CUDA:
> 
> * 0.1.0-jetson8.6.2-rc0001

Here this is technically still trt, just for L4T (Linux for Tegra)

> * 0.1.0-trt24.12-rc0001 (this one works for CPU & GPU?)

That works for both yes. That is, within the containers you can develop/link to cuda etc, but to make use of CUDA at runtime you clearly need a GPU. If your code is not using CUDA, then no problems would arise (but you'd end up with a sizeable image for nothing)

> 
> ML:
> 
> * 0.1.0-jetson8.6.2-rc0001
> * 0.1.0-gpu-trt24.12-rc0001
> * 0.1.0-cpu-trt24.12-rc0001

I am incorrect about on thing in ML. TensorRT base is **only** used for `--target gpu`, otherwise we have a regular ubuntu base where I build some things for CPU.

Which means even my original naming is actually somewhat inaccurate

## new suggestions

- 0.1.0-l4t8.6.2-rc0001 (but more numbers here..)
- 0.1.0-trt24.12-rc0001 

ML: 
- 0.1.0-l4t8.6.2-rc0001
- 0.1.0-trt24.12-rc0001
- 0.1.0-cpu-ub24.04-rc0001 (I guess we reference the true base? I feel like we need a prefix as you said, to make our lives easier with 24.04 and 24.12)

.github/manual-toolkit-push.sh
```diff
@@ -0,0 +1,79 @@
...
+        $SCRIPT_DIR/../toolkits/build.sh --ml-toolkit --target cpu
+      elif [ "$target" = "gpu" ]; then
+        echo "Building for GPU..."
+        $SCRIPT_DIR/../toolkits/build.sh --ml-toolkit --target gpu
+      fi
+    done
+  fi
+
+  # push to registry
+  docker image push --all-tags ghcr.io/aica-technology/toolkits/$type
```
bpapaspyros, line 78: this requires a clean docker image list though. Perhaps it's for as long as the process is manual in any case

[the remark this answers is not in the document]
bpapaspyros: yeah, I actually left a comment for that before (which I resolved for whatever reason).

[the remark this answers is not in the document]
bpapaspyros: so, it keeps the script simpler, but it's potentially dangerous. I can hardcode all the tags (at this point everything is hardcoded anyway)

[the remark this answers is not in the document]
LouisBrunner: Maybe just a confirmation prompt? With a command that shows what will be pushed?

toolkits/build.sh
```diff
@@ -163,19 +168,58 @@ elif [ $ML_TOOLKIT -eq 1 ]; then
...
 BUILD_FLAGS+=(--build-arg=TRT_IMAGE_TAG=${TRT_IMAGE_TAG})
 BUILD_FLAGS+=(--build-arg=VERSION=${VERSION})
-docker buildx build -f Dockerfile."${TYPE}" -t "${IMAGE_NAME}":v"${VERSION}" "${BUILD_FLAGS[@]}" .
+echo "Building image with base \"${IMAGE_NAME}\" and version tags: "
+echo "  - $VERSION"
+for alias in "${ALIASES[@]}"; do
+  echo "  - $alias (alias)"
+done
+
+docker buildx build -f $SCRIPT_DIR/Dockerfile."${TYPE}" -t "${IMAGE_NAME}":v"${VERSION}" "${BUILD_FLAGS[@]}" .
```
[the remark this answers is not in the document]
bpapaspyros: ok, that actually turned out to be a bit more important for manual builds. Since we want multiarch, we need to either build-push the images with `--platform` for both arm and amd, or build the images separately, tag them with their arch and then create a manifest to allow seamless multiarch pulls. 

The latter is what we do with worflows, and I'd prefer it in general, but for ease of use with the manual script I think we need to go with option no1. 

I updated the script accordingly

bpapaspyros:
Ok, our new images are pushed to the registries. I'm taking some time to test and I'll come back here once I'm happy(ish) with what I see.

.github/manual-toolkit-push.sh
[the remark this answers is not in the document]
bpapaspyros: I  changed the logic and commands over here. I am pushing all tags that I built, but it's not baked into the build.sh. So I have --push by default in the workflow, but it really only pushes the new tags and if the build fails then everything fails (which is good)

toolkits/build.sh
```diff
@@ -157,25 +191,89 @@ elif [ $ML_TOOLKIT -eq 1 ]; then
     BUILD_FLAGS+=(--build-arg=TORCHAUDIO_VERSION=$JETSON_TORCHAUDIO_VERSION)
     BUILD_FLAGS+=(--build-arg=TORCHAUDIO_SOURCE=$JETSON_TORCHAUDIO_SOURCE)
     BUILD_FLAGS+=(--build-arg=TARGET=${TARGET})
+    # BUILD_FLAGS+=(--build-arg=CUDA_ARCHS="87;72")
+    BUILD_FLAGS+=(--build-arg=ONNX_BUILD_FOR_GPU=0) # CUDA build is currently failing on tegra TODO: revisit when Microsoft supports this properly
```
bpapaspyros, line 195: sadly, onnxruntime doesn't like tegra for now. I've triex a lot of things but we ultimately fail to compile properly with CUDA. I'm keeping an eye out on their repos, but I don't think this is priority. 

pip dependencies for onnxruntime on tegra (and arm64 in general) also don't exist, so I'm sticking with the manually compiled cpu version. 

I'll put a backlog item to see if we can address this at some point, but it requires a fairly deep dive

bpapaspyros: for python specifically, there are pre-built onnxruntime for Jetpack, but let's keep it for a separate PR. 

We'll still have issues with C++ unless I figure out the compilation though... But having Python is already a good step

bpapaspyros:
small bump here. The images are ready and uploaded in the registries. They work as expected at the moment. More debugging on finer details on what those images may be missing should be handled separately.

bpapaspyros:
> If you could add/amend the tags (if you are happy with the proposal above) then it would be good to go from my side.

sure, I'll correct the script and update the registries right after

toolkits/README.md
```diff
@@ -63,8 +63,8 @@ image = "v4.4.2"
 
 [packages]
 <Other packages you wish to include>
-"@aica/toolkits/cuda-toolkit" = "v0.1.0-24.12-py3"
-"@aica/toolkits/ml-toolkit" = "v0.1.0-gpu-24.12-py3"
+"@aica/toolkits/cuda-toolkit" = "v0.1.0-cuda24.12-rc0001"
+"@aica/toolkits/ml-toolkit" = "v0.1.0-gpu24.12-rc0001"
```
bpapaspyros, line 67:
```suggestion
"@aica/toolkits/cuda" = "v0.1.0-cuda24.12-rc0001"
"@aica/toolkits/ml" = "v0.1.0-gpu24.12-rc0001"
```

toolkits/README.md
```diff
@@ -88,8 +88,8 @@ image = "v2.0.5-jazzy"
 
 # along with the above core libraries, you can also include the toolkit
 # images for easy access to CUDA and ML libraries
-"@aica/toolkits/cuda-toolkit" = "v0.1.0-24.12-py3"
-"@aica/toolkits/ml-toolkit" = "v0.1.0-gpu-24.12-py3"
+"@aica/toolkits/cuda-toolkit" = "v0.1.0-cuda24.12-rc0001"
+"@aica/toolkits/ml-toolkit" = "v0.1.0-gpu24.12-rc0001"
```
bpapaspyros, line 92:
```suggestion
"@aica/toolkits/cuda" = "v0.1.0-cuda24.12-rc0001"
"@aica/toolkits/ml" = "v0.1.0-gpu24.12-rc0001"
```

toolkits/README.md
```diff
@@ -111,7 +111,7 @@ For example, if you specify:
   --target env-vars
 ```
 
-you will obtain a `ghcr.io/aica-technology/cuda-toolkit:vX.Y.Z-24.12-py3` image (refer
+you will obtain a `ghcr.io/aica-technology/cuda-toolkit:vX.Y.Z-cpu24.04-rc0001` image (refer
```
bpapaspyros, line 114:
```suggestion
you will obtain a `ghcr.io/aica-technology/toolkits/cuda:vX.Y.Z-cpu24.04-rc0001` image (refer
```

domire8 (approved):
Nice work with this

PR #156, opened 2025-08-21
feat: add container for yolo model conversions

bpapaspyros:
## Description

- n/a

The idea is to provide an easy to use reference point to be able to convert pytorch (.pt) models to onnx, specifically for [ultralytics]yolo.

This way we can guide our users here instead of asking them to build a virtual env and convert on their own. 

@domire8 if you think this is a bit too much we can remove it. I am not sure it's needed in the registries.

## Review guidelines

## Checklist before merging:


yolo_model_converter/run.sh
```diff
@@ -0,0 +1,32 @@
...
+    exit 1
+    ;;
+  esac
+done
+
+if [ ! -d "$OUTPUT_DIR" ]; then
+  mkdir -p "$OUTPUT_DIR"
+fi
+
+xhost +
```
domire8, line 25:
```suggestion
```

domire8 (commented):
Fair enough, would almost see this as a build-run.sh script directly, easier than having to run to scripts

domire8 (approved):

PR #160, opened 2025-09-01
fix: inject Cuda dependency directly in gpu ml toolkit image

bpapaspyros:
## Description

- #159 

This PR solves the issue by explicitly adding the CUDA toolkit as a dependency.

## Review guidelines

## Checklist before merging:

- [x] Re-test


bpapaspyros:
@LouisBrunner as recommended. I think that works, but I need to take a moment to formally test (which will take a while due to the build), hence draft. I also need to adapt the build script to inject the l4t variant for the arg. 

I am going for a hardcoded version, but perhaps I should be reading the version file from the filesystem and reconstructing the tag (that however, is highly dependent on build.sh arguments, so it's not as straightforward). So I think hardcoding is fine, but adds an extra task when we increase versions

toolkits/Dockerfile.ml
```diff
@@ -296,4 +299,6 @@ ARG VERSION=0.0.0
 LABEL org.opencontainers.image.title="AICA Machine Learning Tool`kit"
 LABEL org.opencontainers.image.description="AICA Machine Learning Toolkit (GPU support)"
 LABEL org.opencontainers.image.version="${VERSION}"
```
LouisBrunner, line 301: I think this should be
```suggestion
LABEL tech.aica.image.metadata='{"type":"lib","dependencies":{\
```

Packages means: "this package also provides the following". This is mostly used by `core` to provide CL and modulo.

Dependencies means: "this package needs those installed as well".

You can check how this is used in our stack on the Registry.

In the case where this should be a "dependency", it should also contain a constraint instead of a version (again the Registry will show you examples), I am imagine you would want something like `~> v0.1.0`.

LouisBrunner (approved):

PR #162, opened 2025-09-08
fix: separate l4t and multiarch versions

bpapaspyros:
## Description

- n/a

I am bumping the versions to v1.0.0 and dropping the rc suffix for multiarch.

L4T images will stay at a separate v0.1.0 version that clearly indicates that they are incompatible with multiarch (there may be uses that will not cause problems, but it will not be by design)

The images have been fairly well tested at this point and I think are a good first release.

That also means we will need dedicated l4t releases in some occasions. Perhaps a future Jetpack version will bring us closer to bridging this version/dependency gap and shipping everything under the same version.

## Review guidelines

Estimated Time of Review: 1 minutes

## Checklist before merging:

### DIFF
--- a/toolkits/README.md
+++ b/toolkits/README.md
@@ -52,7 +52,7 @@ this can be found [here](#building-a-custom-image).
 > may refer to the this [NVIDIA](https://forums.developer.nvidia.com/t/pytorch-for-jetson/72048) forum post.
 >
 > That also means that l4t images can not be used in a drop-in fashion with other multiarch images, as the may
-> contain different library version and/or dependencies.
+> contain different library versions and/or dependencies.
 
 ## Using registry images
 


LouisBrunner (approved):

toolkits/README.md
```diff
@@ -50,6 +50,9 @@ this can be found [here](#building-a-custom-image).
 > 
 > You may also need to install custom versions of PyTorch that are built for Tegra (or compile wheels from source). You
 > may refer to the this [NVIDIA](https://forums.developer.nvidia.com/t/pytorch-for-jetson/72048) forum post.
+>
+> That also means that l4t images can not be used in a drop-in fashion with other multiarch images, as the may
+> contain different library version and/or dependencies.
```
bpapaspyros, line 55:
```suggestion
> contain different library versions and/or dependencies.
```

Issue #185, opened 2026-04-07
Create v3.0.0-jazzy image of ros2-ws


domire8:
The version of ros2 control used in the current ros2-ws image is quite outdated. Due to ros2 control not respecting semver, any update to ros2 control means a breaking change to the ros2-ws image. It is time now to do that and install the most recent version of ros2 control for jazzy.

Additionally, we could implement any of the issues below

@eeberhard @SprGrf @yrh012

SprGrf:
I imagine you mean 3.0.0.

domire8:
@eeberhard any strong opinions about the two sub issues? I feel like Poco and nlohmann-json are fairly common that we could think about installing them in the base

eeberhard:
> I feel like Poco and nlohmann-json are fairly common that we could think about installing them in the base

I agree. The base image does get bigger but it's more important that we have common installations for the needs of control-libraries, modulo and AICA core packages. Given that we already have ROS 2 and ros2_control, it's not that our base image is slim to begin with, so whatever accelerates development and reduces chances of conflicts, the better.

PR #192, opened 2026-04-22
feat: first release candidate for v3.0.0-jazzy

domire8:
## Description

- #185

This PR looks a bit bigger than it is. But since we changed the Dockerfile for 2.0.6 to build on top of 2.0.5, I now have to add back the configuration as we had it for 2.0.5. The main changes are the updated ros2 control tags and the fact that pinocchio is copied over from the newly generated image.

## Review guidelines

## Checklist before merging:


SprGrf (approved):
Builds as expected, let me know on any further tests required.

yrh012 (approved):
looks good to me!

eeberhard (approved):

